PR #15, opened 2018-10-16
State Saver

njalerikson:
Simplified state saving/restoring.

Previously:

```python
idx = self._save_idx()

# do peeking here
...

self._restore_idx(*idx)
```

Now:

```python
with self._src.state:
    # do peeking here
    ...
```

And in cases where the marker also needs to be preserved:

```python
with self._src.state(save_marker=True):
    # do peeking here
    ...
```

A bonus feature that may become valuable for number versus date parsing is the ability to save multiple states:

```python
with self._src.state:
    # do some peeking here
    ...
    with self._src.state:
        # do some more peeking here
        ...
```

### DIFF
--- a/tomlkit/parser.py
+++ b/tomlkit/parser.py
@@ -66,11 +66,11 @@ class _State:
         self.restore()
 
     def save(self):  # type: () -> None
-        self._index = self._src._index
+        self._idx = self._src._idx
         self._marker = self._src._marker
 
     def restore(self):  # type: () -> None
-        self._src.index = self._index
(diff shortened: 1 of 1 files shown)


njalerikson:
You can see at these lines in `_parse_value` how I expect this state saver to be utilized.
https://github.com/sdispater/tomlkit/blob/13df55136bfe382aa29c365591799cff83567f27/tomlkit/parser.py#L631-L635

njalerikson:
@sdispater anything I can do to get this rolling?

PR #22, opened 2018-10-24
Update boolean parsing

njalerikson:
Switch boolean parsing to also be a character by character parser using the state saving logic.


tomlkit/parser.py
```diff
@@ -466,27 +473,30 @@ def _parse_value(self):  # type: () -> Item
         """
         Attempts to parse a value at the current position.
         """
-        self.mark()
+        try:
+            return self._is_string(StringType.SLB)
```
sdispater: I am not sure about that. I would expect a `is_xxx()` method to return a boolean and not to execute an action.

I think the previous way of doing is much clearer.

```python
if c == StringStype.SLB.value:
    self._parse_basic_string()
```

That's what I did here https://github.com/sdispater/tomlkit/commit/428a9b2c4f785bd17e81b2940383a3dc7bd6bed2#diff-58b7a35cabb00a194ef1dc33829a5975R473

You will see that I also removed the need of raising a `Restore` exception since it makes the code more complex than it should be.

njalerikson:
But at the same time the changes that have been made eliminating the Restore logic effectively eliminates any value in using the state at all inside of parse value. After the first conditional check we will actually never restore and try a different value parser.

The Restore logic previously envisioned would be critical for #26 where I first try parsing the value as a datetime, when that fails we restore and try parsing for a date, when that fails we restore and try parsing for a time, etc.

sdispater:
For #26 I don't think it warrants the use of a state. The current way works well:

- If the current value starts with a digit or `+` or `-` it must be a number or a date/datetime/time.
- We extract the value and inspect it to determine its type.
- If we can't determine the type, it's an error and we raise it.

PR #23, opened 2018-10-24
Moved array parsing

### LINKED ISSUE #25: Array's commas
In accordance with the grammar 

https://github.com/toml-lang/toml/blob/bb47759841ac368d86eb7a459bd7eea7162b9a80/toml.abnf#L188-L200

leading and duplicate commas are not allowed (trailing is allowed).

In other words the following cases should not be allowed (but currently are possible):

1. **No comma**:
    ```
    a = [ 12  34 ]
    ```

1. **Leading comma**:
    ```
    a = [ , 12 ]
    ```

1. **Duplicate comma**:
    ```
    a = [ 12 ,,,,,, 34 ]
    ```

njalerikson:
Moved array parsing into self method. ~No actual code change.~

Fixes #25.


tomlkit/parser.py
```diff
@@ -474,6 +474,9 @@ def _parse_value(self):  # type: () -> Item
         with self._state:
             return self._parse_literal_string()
 
+        with self._state:
```
sdispater: Initializing a state before even knowing if we have a potential array will impact performance.

I think we can still have the array parsing in its own method but the character check should still be in `_parse_value()`.

njalerikson: I agree I've seen a slight slowdown. My ultimate desire though was to get `_parse_value` to look like this:

```python
def _parse_value(self):
    for parser in self._value_parsers:
        with self.state:
            return parser()
```

where `self._value_parsers` is the various parsing methods. Maybe it should look something more like this then:

```python
def _parse_value(self):
    for parser in self._value_parsers:
        try:
            return parser()
        except Restore:
            pass
```

where a `self._value_parsers` looks like this:

```python
def _parse_array(self):
    if self._current != "[":
        raise Restore

    with self.state:
        ...
```

That way we are still moving towards modularity but also have the ability to exit a scope before a state is made?

njalerikson:

PR #24, opened 2018-10-24
Moved inline table parsing

### LINKED ISSUE #19: Inline Table's commas
In accordance with the grammar 

https://github.com/toml-lang/toml/blob/bb47759841ac368d86eb7a459bd7eea7162b9a80/toml.abnf#L213-L221

leading, trailing, and duplicate commas are not allowed.

In other words the following cases should not be allowed (but currently are possible):

1. **No comma**:
    ```
    a = { b = 12  c= 'hello' }
    ```

1. **Leading comma**:
    ```
    a = { , b = 12 }
    ```

1. **Duplicate comma**:
    ```
    a = { b = 12 ,,,,,, c='hello' }
    ```

1. **Trailing comma**:
    ```
    a = { b = 12 , }
    ```

~I have also addressed this with a fix dependent upon #15.~

njalerikson:
Moved inline table parsing into its own method. Also fixed up the parsing to properly address commas. Applicable unittests have been added.

Fixed #19.


njalerikson:

PR #26, opened 2018-10-24
Split number from date parsing

njalerikson:
Converted number and date parsing into character by character parsing along with relying on the state saving logic.


njalerikson:

PR #31, opened 2018-11-09
Trim comments when setting InlineTable

### LINKED ISSUE #30: Convert an table to inline table, the comment is preserved unexpectedly
Suppose we have following toml:
```toml
[site.user]
name = "John"  # Inline comment
age = 28
```
I want to convert `site.user` to an inline table:
```python
v = parsed['site']['user']
table = tomlkit.inline_table()
table.update(v)
print(table.as_string())
```
```toml
{age = 21,name = "John"# Inline comment}
```
What is weird is that even with:
```python
table.update(dict(v))
```
The bug still exists.

The expected behavior should be dropping all comments of original table.

frostming:
InlineTable is supposed to be one liner and doesn't support comments in values.

Closes #30

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -424,5 +424,7 @@ def test_trim_comments_when_building_inline_table():
     row = parse('foo = "bar"  # Comment')
     table.update(row)
     assert table.as_string() == '{foo = "bar"}'
-    table.append("baz", row["foo"])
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks!

Looks good to me 👍

techalchemy:
Thanks for the quick responses and for putting together a good toml library for python, it was badly needed

PR #34, opened 2018-11-28
Pick the max when there are many indices for a key.

frostming:
The test case demonstrates how it fails when appending a table.


sdispater:
Thanks!

I will try to make a bugfix release as soon as possible.

Issue #40, opened 2019-03-06
Subtracting two dates incorrectly return a Date object


dunkmann00:
When you subtract two Date objects tomlkit takes the result and makes it into a new Date object:
```python
def __sub__(self, other):
        result = super(Date, self).__sub__(other)

        return self._new(result)
```

This behavior is incorrect as the value that should get returned is a `timedelta` object as per the [Python Docs](https://docs.python.org/3/library/datetime.html#date-objects). 

The behavior is implemented correctly for `Datetime` objects in tomlkit, so I'll go ahead and make a pull request with the changes added to the `Date` object as well.

PR #41, opened 2019-03-06
When subtracting two Date objects, a timedelta object is returned.

dunkmann00:
Here is the pull request that addresses #40. Let me know if there is anything else you need me to do!


sdispater:
Thanks!

Issue #49, opened 2019-05-03
setdefault does not behave as expected


AndydeCleyre:
> [TOMLDocument] behaves like a standard dictionary

```python
>>> from tomlkit import parse
>>> tomldoc = parse('''[table]\nfoo="bar"''')
>>> tomldoc
{'table': {'foo': 'bar'}}
>>> data = {'table': {'foo': 'bar'}}
>>> tomldoc['table'].setdefault('baz', 'waldo')
'waldo'
>>> data['table'].setdefault('baz', 'waldo')
'waldo'
>>> tomldoc
{'table': {'foo': 'bar'}}
>>> data
{'table': {'foo': 'bar', 'baz': 'waldo'}}
```

I expected the `TOMLDocument` to gain a `baz` entry after the `setdefault` call. This would be helpful for adding data to document sections which may or may not exist yet.

```python
tomlkit==0.5.3
```

AndydeCleyre:
I'll add here the use case that made me aware of this issue: scripting the insertion of requirements from pip-tools style `requirements.in` files into a flit style `pyproject.toml` file.

```python
#!/usr/bin/env python3
from plumbum import local
import tomlkit


suffix = 'requirements.in'
pyproject = local.cwd.up() / 'pyproject.toml'


if __name__ == '__main__':
    if pyproject.is_file():
        toml_data = tomlkit.parse(pyproject.read())
        for reqsin in local.cwd // f'*{suffix}':
            pyproject_reqs = [
                line
                for line in reqsin.read().splitlines()
                if line.strip() and not line.startswith('#')
            ]
            extras_catg = reqsin.name.rsplit(suffix, 1)[0].rstrip('-.')
            if not extras_catg:
                toml_data['tool']['flit']['metadata']['requires'] = pyproject_reqs
            else:
                # toml_data['tool']['flit']['metadata'].setdefault('requires-extra', {})
                if 'requires-extra' not in toml_data['tool']['flit']['metadata']:  #
                    toml_data['tool']['flit']['metadata']['requires-extra'] = {}  #
                toml_data['tool']['flit']['metadata']['requires-extra'][extras_catg] = pyproject_reqs
        pyproject.write(tomlkit.dumps(toml_data))
```

It seems to me that the commented line ought to be able to replace the two following lines.

AndydeCleyre:
Please have a look at my pull request which fixes the issue as I see it, with a corresponding test.

AndydeCleyre:
[...]

PR #50, opened 2019-05-04
Bugfix/49

AndydeCleyre:
Simple logic and test for the dict-like setdefault behavior discussed in #49 .

### DIFF
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -103,6 +103,25 @@ bar = "boom"
 """
     )
 
+    assert doc.setdefault("bar", "waldo") == "boom"
+
+    assert (
+        doc.as_string()
+        == """foo = "bar"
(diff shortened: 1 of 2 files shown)


Julian:
Just found this after encountering the same bug -- something seems off since at this point the PR doesn't look like it's touching the test suite, the test suite on master passes, but the bug persists, so perhaps something went off in one of the merges, I didn't investigate.

Personally, I'd probably try to fix this "for good" by removing the inheritance from builtin classes, which is IME often the root cause of these sorts of bugs, but certainly any fix is an improvement

AndydeCleyre:
@sdispater 

I just had another look at this today and pushed an updated attempt, which looks to me successful. Please review when you can, thanks!

frostming:
Already fixed in the master

PR #57, opened 2019-07-23
Let items.Bool inherit from long to get comparison

### LINKED ISSUE #56: Boolean value in document causes erratic comparison results
Via sarugaku/plette#9.

```pycon
>>> import tomlkit
>>> s = """
... [foo]
... value = false
... """
>>> tomlkit.parse(s) == {'foo': {'value': False}}    # This works correctly.
True
>>> tomlkit.parse(s)['foo'] == {'value': False}      # This does not.
False
```

At a quick glance, `tomlkit.items.Bool` seems to be the problem here.

```pycon
>>> d = tomlkit.parse(s)
>>> d
{'foo': [{'value': False}]}
>>> d['foo']
{'value': <tomlkit.items.Bool object at 0x10c9ac780>}
```

uranusjr:
Attempt to fix #56.

### DIFF
--- a/tomlkit/items.py
+++ b/tomlkit/items.py
@@ -491,10 +491,13 @@ class Float(float, Item):
         return float(self), self._trivia, self._raw
 
 
-class Bool(Item):
+class Bool(Item, long):
     """
     A boolean literal.
     """
+
+    def __new__(cls, t, trivia):  # type: (float, Trivia) -> long
+        return long.__new__(cls, bool(t))
 
     def __init__(self, t, trivia):  # type: (float, Trivia) -> None
         super(Bool, self).__init__(trivia)


frostming:
Any updates on this issue? This also affects `isinstance` check, which is used by data validator. So I prefer to inherit from `bool`

sdispater:
@frostming In Python, it's impossible to inherit from the `bool` type. If you try you will get:

```
TypeError: type 'bool' is not an acceptable base type
```

I addressed the problem as much as possible in this commit 4636463553d83c32b2a55505806dd319f2969ddf but there isn't any way to make `isinstance()` or `doc["my-boolean-value"] is True` work, unfortunately.

Anyway, since the previously mentioned commit fixed what was fixable, i am closing this PR.

PR #60, opened 2019-08-23
inline tables should have newlines after them

toppk:
I noticed after adding and removing some dependencies with extras, that newlines were disappaering in my pyproject.toml.  I tracked it down to inline_tables not including newlines.  I'm not really familiar with the codebase, so this fix may leave other gaps  but I think it addresses a common pattern.

included are:

  -proposed fix
  -test case

You can see a realworld example of this bug by running this sequence a few times on a healthy pyproject.toml file

poetry add  -E bcrypt -E argon2 passlib
poetry remove  passlib

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -12,6 +12,7 @@ from datetime import timedelta
 
 from tomlkit import inline_table
 from tomlkit import parse
+from tomlkit import dumps
 from tomlkit._compat import PY2
 from tomlkit.exceptions import NonExistentKey
 from tomlkit.items import InlineTable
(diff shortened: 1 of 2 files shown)


toppk:
check failures unrelated to my patch :angry:

toppk:
regarding the pypy failure, it seems that the preview version of poetry requires versions of the typing module that pypy doesn't have in any release or beta version.  I've created an issue over there.
(for reference)
https://github.com/sdispater/poetry/issues/1319

sdispater:
Thanks for your contribution!

However, I think this already fixed in the latest release.

Issue #80, opened 2020-02-29
.get on class OutOfOrderTableProxy now always return None


mbelang:
I just updated to poetry 1.0.4 that uses tomlkit 0.5.10 and custom repository are not added anymore.

```
(Pdb++) local_config.get("source")
(Pdb++) local_config.get("name")
(Pdb++) local_config.get("name", "potatoes")
'potatoes'
(Pdb++) local_config["name"]
'test'
(Pdb++) local_config["source"]
<AoT [{'name': 'testrepo', 'url': 'https://test-repo-url'}]>
```

I'm pretty sure it is related to that change: https://github.com/sdispater/tomlkit/pull/79

OS: ArchLinux
Python: 3.7, 3.8

Reproduced in docker as well with images python:3.7.x and python:3.8.x

PR #81, opened 2020-02-29
fix: implement get on out of order container proxy

mbelang:
fixes: https://github.com/sdispater/tomlkit/issues/80


mbelang:
@sdispater please :)

PR #91, opened 2020-04-03
Allow cleaner TOMLDocument.get() usage

Haeilifax:
This allows calling TOMLDocument.get() on dotted keys. For example, if your document was:

```
"""
[content]
key1 = "value"
    [content.table1]
        "key1.1" = "value2"
"""
```

Then, once it's parsed into a TOMLDocument (call it doc), you can get "value2" through

`doc.get("content.table1.'key1.1'") `

This method does not preclude you from still doing, say,

`doc["content"]["table1"]["key1.1"]`

or

`doc.get("content").get("table1").get("'key1.1'")`

This new functionality also allows you to use defaults on a whole chain of get commands - if you've parsed a document and aren't sure if it contains a specific key (or even any keys at all), you can do

`doc.get("content.table1.thiskeydoesnotexist", "Safe Value")`

This will return "Safe Value" if any of the dotted keys (content, table1, or thiskeydoesnotexist) do not exist. Also included are tests for the new functionality. Do note a change in the get signature for quoted keys with dots in them - you must wrap them in extra quotes, or they will be parsed as a dotted key.

(Old pull request was deleted to use non-master branch)


ptapping:
This would be a great feature to include. I only just started using tomlkit and already sick of typing [""][""][""]...

frostming:
`content.table1.'key1.1'` is a valid key itself. Adding this will cause confusion.

PR #99, opened 2020-05-27
Add flake8 and isort

bobfang1992:
Hi,

There is no functional change at all. This is simply adding `flake8` and `isort` to this project to make it easier to manage.

Please let me know if you have any questions.

Bob

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,18 @@
+[flake8]
+max-line-length = 81
+ignore = E501, E203, W503, F821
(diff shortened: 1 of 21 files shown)


bobfang1992:
I am also considering add mypy support but that is quite a big change so not going to include in this PR.

bobfang1992:
Why closed without any reason?

frostming:
They are already added to the project

PR #108, opened 2020-09-16
Add support for toml files in UTF8 with BOM

duxiaoyao:
Throw tomlkit.exceptions.EmptyKeyError from TOMLFile.read for a toml file in UTF8 with BOM. 

BTW, UTF8 file in Windows is with BOM by default.


duxiaoyao:
Don't know how to run build local to reproduce the fail. Need for help here. @sdispater

frostming:
Quote from spec: https://toml.io/en/v1.0.0
> A TOML file must be a valid UTF-8 encoded Unicode document.

PR #116, opened 2020-12-25
Fix incorrect InvalidControlChar error message

Qwaz:
TOML `\u` unicode literal takes a hex string, but the error message was using `str()`. This results in an incorrect suggestion to users.


frostming:
LGTM

PR #120, opened 2021-04-30
Remove "tests" package from sdist installation

### LINKED ISSUE #69: setup.py installs the tests package
This might be partly a bug in poetry but the sdist tarball's setup.py installs the tests package alongside the tomlkit package.

kwshi:
Resolves #69 (nice).  Currently the tomlkit tests are included in the distribution via `packages`, but that behavior is incorrect since it causes `tests` to be exposed (and installed!) as a module.  We don't want that; we only want `tests` to be _included_ in the source distribution, so the more sensible configuration is to use the `include` key.

Solution comes from [this comment](https://github.com/python-poetry/poetry/issues/2353#issuecomment-578438198), as well as [@abn's implementation of the same fix for Poetry](https://github.com/python-poetry/poetry/pull/4007).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,8 +12,9 @@ repository = "https://github.com/sdispater/tomlkit"
 
 packages = [
     {include = "tomlkit"},
-    {include = "tests", format = "sdist"}
 ]
+
+include = ["tests"]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.5"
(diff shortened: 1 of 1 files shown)


eli-schwartz:
It is a teeny bit disappointing this fix did not get merged in time for 0.7.1, released 2 hours ago. Hopefully it can be merged in time for any future 0.7.2?

kwshi:
For reference, this PR was closed because I think the issue was fixed in 63b96f11456e987a9c54267e9d8020acc5d4cc9d.

Issue #130, opened 2021-05-25
Public API seems to rely on a private type (type hints)


abravalheri:
Hello, thank you very much for the amazing lib.

Lately I have been working in a project that uses `tomlkit` and relies on mypy for the typechecking.

Unfortunately, it seems to me that both `tomlkit.dumps` ans `tomlkit.loads` type signatures depend on what seems to be a private class... This is specially suggested by: 

https://github.com/sdispater/tomlkit/blob/a4d43a772490f28b16bc2964888a103783de3a0b/tomlkit/api.py#L25

Here the class is imported as `_TOMLDocument` which seems to imply that it is an implementation detail, and private.

The problem arises when trying to use those functions in a type-safe manner, since users are not supposed to import private terms and `TOMLDocument` is not exposed in `__init__.py`...

Moreover the implementation of `tomlkit.dumps` suggests that the type hint is not accurate, I suppose the correct would be `(Union[TOMLDocument, dict], bool) -> str`?

PR #131, opened 2021-05-25
Expose public class in type hints for dumps/loads (#130)

abravalheri:
Hello, this is an attempt to solve #130.

Unfortunately, I had lots of trouble trying to run the test suite locally (with both `tox` and ``poetry run pytest tests/`), so sorry for advance if something goes wrong (what I ended up doing was firing up a python terminal, loading the `test_api` file and manually running `test_mapping_types_can_be_dumped()`.

### DIFF
--- a/tomlkit/api.py
+++ b/tomlkit/api.py
@@ -2,6 +2,7 @@ import datetime as _datetime
 
 from typing import Tuple
 
+from ._compat import Mapping
 from ._utils import parse_rfc3339
 from .container import Container
 from .items import AoT
@@ -22,10 +23,10 @@ from .items import Trivia
 from .items import Whitespace
 from .items import item
 from .parser import Parser
-from .toml_document import TOMLDocument as _TOMLDocument
+from .toml_document import TOMLDocument
 
 
-def loads(string):  # type: (str) -> _TOMLDocument
+def loads(string):  # type: (str) -> TOMLDocument
     """
     Parses a string into a TOMLDocument.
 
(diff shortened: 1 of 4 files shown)


tomlkit/api.py
```diff
@@ -22,10 +23,10 @@
 from .items import Whitespace
 from .items import item
 from .parser import Parser
-from .toml_document import TOMLDocument as _TOMLDocument
```
abravalheri, line 25: The type hints of public functions (e.g. loads/dumps) where exposing something that looked a private class...
This makes it complicated to use the library in a typechecked project.

tomlkit/api.py
```diff
@@ -34,28 +35,28 @@ def loads(string):  # type: (str) -> _TOMLDocument
...
 
 
-def dumps(data, sort_keys=False):  # type: (_TOMLDocument, bool) -> str
+def dumps(data, sort_keys=False):  # type: (Mapping, bool) -> str
     """
     Dumps a TOMLDocument into a string.
     """
-    if not isinstance(data, _TOMLDocument) and isinstance(data, dict):
-        data = item(data, _sort_keys=sort_keys)
+    if not isinstance(data, Container) and isinstance(data, Mapping):
```
abravalheri, line 42: I changed `TOMLDocument` to `Container` here, because I imagine that structures that are already internal to tomlkit are supposed to rely on `as_string` and that might be the case some of them are already `Mapping`s (maybe TOML tables?).

The changes here are indeed making the function *more acceptive* in terms of arguments, e.g. `MappingProxy` (a Python idiom that makes it easy to pass "immutable dicts" around), `OrderedDict` (although probably the order will not be preserved for Python < 3.6). Alternatively we could also use `Union[dict, TOMLDocument]` to avoid augmenting the function's domain.

PR #149, opened 2021-11-17
Prevents TOMLFile from changing line endings

utek:
I think that line endings should be preserved and not changed
to system ones when reading and saving the file.


abn:
@sdispater can you review this please?

Relates-to: https://github.com/python-poetry/poetry/issues/2993

utek:
@sdispater I've modified the poetry install in the workflow -> https://github.com/sdispater/tomlkit/pull/149/commits/359e9c41fb49a334754729cecc49fc8a370ffeac I think the preview version of poetry can't be installed this way (according to error msg)

.github/workflows/tests.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         shell: bash
         run: |
           curl -fsS -o get-poetry.py https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py
-          python get-poetry.py --preview -y
+          python get-poetry.py -y
```
abn, line 55: You can port over to the new script now. https://github.com/python-poetry/poetry/blob/9e31f7a2f040f274148fec8fc6aae3721d2e6a90/.github/workflows/main.yml#L53-L66 `--preview` is no longer required i think.

frostming:
Thanks

utek:
@frostming Thank you

PR #150, opened 2021-11-18
Update pre-commit configuration with the up-to-date urls and versions

utek:


frostming:
Thanks, this is superseded in #151

PR #163, opened 2021-12-24
Initialize Sphinx docs

frostming:


honno:
https://tomlkit.readthedocs.io/en/latest/quickstart.html isn't live for me :confused:

frostming:
Let's wait for the RTFD hooks set up. I don't have enough permission right now.

PR #168, opened 2022-01-27
Fix values similar to booleans being incorrectly parsed to booleans

### LINKED ISSUE #165: Bug: tomlkit.value("test") returns True
Since tomlkit `0.8.0` all strings starting with lowercase letter `t` are converted to `True` and all strings starting with `f` are coverted to `False` if using the `tomlkit.value(...)` call.

E.g.: Install tomlkit `0.8.0` into a fresh venv and using `python3`:
```sh
❯ python3                   
Python 3.10.0 (default, Oct 22 2021, 13:24:07) [Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import tomlkit
>>> tomlkit.value("test")
True
>>> tomlkit.value("fest")
False
>>> tomlkit.value("NoTest")
  File "<stdin>", line 1, in <module>
  File "/Users/robot/test/test123/lib/python3.10/site-packages/tomlkit/api.py", line 149, in value
    return Parser(raw)._parse_value()
  File "/Users/robot/test/test123/lib/python3.10/site-packages/tomlkit/parser.py", line 598, in _parse_value
    raise self.parse_error(UnexpectedCharError, c)
tomlkit.exceptions.UnexpectedCharError: Unexpected character: 'N' at line 1 col 0
```

**I would expect that all these strings throw an error like it is done in `0.7.2`.**

MrGreenTea:
This included values that strat with `t` or `f` like `test` and `friend` as well as values that have `true` or `false` as prefix like `truely` and `falsely`.

All of them are now incorrectly rejected by `tomlkit.value`.

Closes #165

### DIFF
--- a/tomlkit/parser.py
+++ b/tomlkit/parser.py
@@ -424,9 +424,16 @@ class Parser:
             return self._parse_basic_string()
         elif c == StringType.SLL.value:
             return self._parse_literal_string()
-        elif c == BoolType.TRUE.value[0]:
+        # peek one furhter then the value so we don't parse values that look like trueXXX or falseXXX
+        elif (
+            c == BoolType.TRUE.value[0]
+            and self._peek(len(BoolType.TRUE.value) + 1) == BoolType.TRUE.value
+        ):
             return self._parse_true()
-        elif c == BoolType.FALSE.value[0]:
+        elif (
+            c == BoolType.FALSE.value[0]
+            and self._peek(len(BoolType.FALSE.value) + 1) == BoolType.FALSE.value
+        ):
             return self._parse_false()
         elif c == "[":
             return self._parse_array()
@@ -1082,7 +1089,7 @@ class Parser:
         with self._state(restore=True):
             buf = ""
             for _ in range(n):
-                if self._current not in " \t\n\r#,]}":
+                if self._current not in " \t\n\r#,]}" + self._src.EOF:
                     buf += self._current
                     self.inc()
                     continue
--- a/tomlkit/parser.py
+++ b/tomlkit/parser.py
@@ -123,11 +123,11 @@ class Parser:
         """
         self._src.mark()
 
-    def parse_error(self, exception=ParseError, *args):
+    def parse_error(self, exception=ParseError, *args, **kwargs):
         """
         Creates a generic "parse error" at the current position.
         """
-        return self._src.parse_error(exception, *args)
+        return self._src.parse_error(exception, *args, **kwargs)
 
     def parse(self) -> TOMLDocument:
         body = TOMLDocument(True)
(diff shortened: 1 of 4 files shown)


tomlkit/parser.py
```diff
@@ -424,9 +424,10 @@ def _parse_value(self) -> Item:
             return self._parse_basic_string()
         elif c == StringType.SLL.value:
             return self._parse_literal_string()
-        elif c == BoolType.TRUE.value[0]:
+        # peek one furhter then the value so we don't parse values that look like trueXXX or falseXXX
+        elif BoolType.TRUE.value.startswith(c) and self._peek(len(BoolType.TRUE.value) + 1) == BoolType.TRUE.value:
```
frostming, line 428: Is `startswith` more efficient than `==` ? any other reason for changing this?

MrGreenTea: I just find it more clear what the actual check is. The check could also just be completely removed, as it's included in the second term

frostming: Yes, but I would prefer to keep it as a shortcut.

frostming: Also remember to format the code with `black` before commit. It is integrated into `pre-commit` if you have installed it.

MrGreenTea: Yeah sorry, I overlooked installing the `pre-commit` hooks. I've fixed those issues now and reverted the check to a simple equality `==` check

frostming:
Sorry, I revisited the original issue and fix, I come up with a different fix approach.

The real problem here are 1) the error [isn't thrown correctly](https://github.com/sdispater/tomlkit/commit/3dcd42fcb125cf9df51981bace2ac3a3e06a1b38) and 2) the `value()` API doesn't check the EOF of the given string.

The parsing of boolean values is correct. So can you kindly revert the change in parsing but ensure the EOF is reached after [value parsing](https://github.com/sdispater/tomlkit/commit/a304a39071cef76886e472be6c89d44374a8ba1a). Thanks for your patience.

MrGreenTea:
I'm not sure I follow exactly. Can you provide me with some tests that reproduce the behaviour you expect? Or some examples?

frostming:
```python
tomlkit.value('"foo"1.2')
# should reject
```
the current behavior parses as 'foo' but it doesn't reach the end. `1.2` is remaining unparsed, we should check against this case. Then the original issue will also be fixed.

MrGreenTea:
If I understand you correctly, we should check if EOF was reached when parsing a value?
Would a simple check like this suffice?
```py
if not self.end():
    raise ParseError
```

frostming:
lgtm, but it only applies to value api so it should be parser.end()

also a meaningful error message would be good

frostming:
LGTM! Thanks

MrGreenTea:
@frostming thanks for the help! Is there a chance to get a new release with this bugfix soon?

PR #177, opened 2022-02-10
Allow string type to be controlled from the public API

### LINKED ISSUE #44: Multiline dump?
If a string contains `\n`, is it possible to dump as multiline? e.g
```
foo = """
my 
string
"""
```

abravalheri:
Hello, it would be interesting to expose a way of controlling the string style (e.g. allowing multi line strings).

This PR is an attempt to provide such API.

I ended up going for a very verbose API (with explicit keywords) because I believe that it is easier to use, but please let me know if you prefer exposing `StringType` as part of the public API.

I think this closes #44.

### DIFF
--- a/tomlkit/api.py
+++ b/tomlkit/api.py
@@ -110,7 +110,7 @@ def string(
     *,
     single_quotes: bool = False,
     multiline: bool = False,
-    escape: Union[None, bool] = None
+    escape: Union[None, bool] = None,
 ) -> String:
     """Create a string item.
 
@@ -120,7 +120,7 @@ def string(
     By default, single line strings are escaped, but multi line strings are not.
     This can be controlled by explicitly setting ``escape``.
     """
-    escape = not(multiline) if escape is None else escape
+    escape = (not multiline) if escape is None else escape
     type_ = _StringType.select(single_quotes, multiline)
     return String.from_raw(raw, type_, escape)
 
--- a/tomlkit/api.py
+++ b/tomlkit/api.py
@@ -23,6 +23,7 @@ from .items import Item as _Item
 from .items import Key
 from .items import SingleKey
 from .items import String
+from .items import StringType as _StringType
 from .items import Table
 from .items import Time
 from .items import Trivia
@@ -104,9 +105,24 @@ def boolean(raw: str) -> Bool:
     return item(raw == "true")
 
 
-def string(raw: str) -> String:
-    """Create a string item."""
-    return item(raw)
+def string(
+    raw: str,
+    *,
+    single_quotes: bool = False,
+    multiline: bool = False,
+    escape: Union[None, bool] = None
+) -> String:
+    """Create a string item.
+
+    Boolean flags (e.g. ``single_quotes=True`` and/or ``multiline=True``)
+    can be used for personalization.
+
+    By default, single line strings are escaped, but multi line strings are not.
+    This can be controlled by explicitly setting ``escape``.
+    """
+    escape = not(multiline) if escape is None else escape
+    type_ = _StringType.select(single_quotes, multiline)
+    return String.from_raw(raw, type_, escape)
 
 
 def date(raw: str) -> Date:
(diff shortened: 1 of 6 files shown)


tomlkit/api.py
```diff
@@ -104,9 +105,24 @@ def boolean(raw: str) -> Bool:
     return item(raw == "true")
 
 
-def string(raw: str) -> String:
-    """Create a string item."""
-    return item(raw)
+def string(
+    raw: str,
+    *,
+    single_quotes: bool = False,
```
frostming, line 111: I suggest using the TOML term: literal(`'`) or bare(`"`)

tomlkit/api.py
```diff
@@ -104,9 +105,24 @@ def boolean(raw: str) -> Bool:
     return item(raw == "true")
 
 
-def string(raw: str) -> String:
-    """Create a string item."""
-    return item(raw)
+def string(
+    raw: str,
+    *,
```
frostming, line 110: I am happy this is keyword-only

tomlkit/items.py
```diff
@@ -1512,6 +1519,12 @@ def _new(self, result):
     def _getstate(self, protocol=3):
         return self._t, str(self), self._original, self._trivia
 
+    @classmethod
+    def from_raw(cls, value: str, type_=StringType.SLB, escape=False) -> "String":
+        string_value = escape_string(value) if escape else value
```
frostming, line 1524: `escape_string()` needs to take the string type as a parameter to decide whether a quote should be escaped or not.

abravalheri: I was revisiting the standard and It seems that escaping is more complicated than that :/

I did some changes to the implementation of `escape_string` to be parameterised in terms of `escape_chars` and `escape_sequences`, and then I defined both groups for each string type.

Hopefully it works better now? Please let me know what you think!

abravalheri:
Thank you very much for the kind review @frostming.

I changed the implementation to address escaping properly trying to follow what the standard has to say about which characters and sequences that can or cannot appear in the string value...

(P.S.: I had one doubt about the standard text, so I opened this issue: https://github.com/toml-lang/toml/issues/877)

tomlkit/items.py
frostming: I'm sorry I may make a mistake here. According to toml's spec, there is no escaping in literal strings. Therefore, `'\t'(python) != '\t'(toml)`, and:
>Since there is no escaping, there is no way to write a single quote inside a literal string enclosed by single quotes.

If we are to build a string with `string("'", literal=True, multiline=False)`, it should raise an error.

abravalheri: Yes: `'\t'(python) != '\t'(toml)`, this is correct. `'\t' (toml) == r'\t' (python).

So we have 2 alternatives here:

a) Raise an error when invalid characters are added with `literal=True`
b) Make the string valid... (i.e. be lenient with the input but strict with the output)

Since the `escape_string` function is not public, I implemented (b) because I think it is easier to use. If we consider that the user can still pass `escape=False`, this approach is the one with maximum flexibility.

I tried to describe it in the docstring. Hopefully is easier to understand?

Do you think that it is better to fail instead?

frostming: I prefer solution a)

It is sometimes not possible to make the string valid when `literal=True`. So the whole logic would be:

1. `literal=False, multiline=False`. The escape characters are the MUST escapes plus `\t` and `"`(Yes, we prefer to escape `\t` for better looking)
2. `literal=False, multiline=True`. The escape sequences are the MUST escapes plus `\t` and `"""`
3. `literal=True, multiline=False`. No escapes, raise an error if `'` is in the string.
4. `literal=True, multiline=False`. No escapes, raise an error if `'''` is in the string.

abravalheri: Thank you.

Just submitted a new patch with the method failing for invalid characters as discussed. Please let me know your thoughts.

PR #187, opened 2022-04-12
Add unwrap method to recursive convert to plain old python objects

### LINKED ISSUE #43: Function to convert everything to plain old python objects
As more projects incorporate PEP518, many will be aiming to keep compatibility with their existing config files as well (be they ini, yaml, json, etc.). In order to help abstract over config files of different types, it would be useful to be able to convert tomlkit objects into POPOs. One of TOML's strengths for python is that all of its constructs can deserialise to POPOs: could this be made available for tomlkit users?

syntapy:
Adds methods to convert TOMLKit objects to plain python methods

Fixes #43

### DIFF
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -48,7 +48,7 @@ class Container(_CustomDict):
 
     def unwrap(self, recursive: bool = True) -> str:
         unwrapped = {}
-        for k, v in self._body:
+        for k, v in self.items():
             if k is None:
                 continue
 
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -46,22 +46,25 @@ class Container(_CustomDict):
     def body(self) -> List[Tuple[Optional[Key], Item]]:
         return self._body
 
+    def _unwrap_inner(self, k, v, d):
+        if k is None:
+            return
+
+        k = k.key
+        v = v.value
+
+        if isinstance(v, Container):
+            v = v.value
+
+        if k in d:
+            merge_dicts(d[k], v)
+        else:
+            d[k] = v
+
     def unwrap(self) -> str:
         unwrapped = {}
         for k, v in self.items():
-            if k is None:
-                continue
-
-            k = k.key
-            v = v.value
-
-            if isinstance(v, Container):
-                v = v.unwrap()
-
-            if k in unwrapped:
-                merge_dicts(unwrapped[k], v)
-            else:
-                unwrapped[k] = v.unwrap()
+            self._unwrap_inner(k, v, unwrapped)
 
         return unwrapped
 
@@ -69,19 +72,7 @@ class Container(_CustomDict):
     def value(self) -> Dict[Any, Any]:
         d = {}
         for k, v in self._body:
-            if k is None:
-                continue
-
-            k = k.key
-            v = v.value
-
-            if isinstance(v, Container):
-                v = v.value
-
-            if k in d:
-                merge_dicts(d[k], v)
-            else:
-                d[k] = v
+            self._unwrap_inner(k, v, d)
 
         return d
 
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -10,6 +10,7 @@ from typing import Union
 
 from ._compat import decode
 from ._utils import merge_dicts
+from .check import is_tomlkit
 from .exceptions import KeyAlreadyPresent
 from .exceptions import NonExistentKey
 from .exceptions import TOMLKitError
@@ -24,8 +25,6 @@ from .items import Trivia
 from .items import Whitespace
 from .items import _CustomDict
 from .items import item as _item
-
-from .check import is_tomlkit
 
 
 _NOT_SET = object()
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -46,6 +46,25 @@ class Container(_CustomDict):
     def body(self) -> List[Tuple[Optional[Key], Item]]:
         return self._body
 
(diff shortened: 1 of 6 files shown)


tomlkit/container.py
```diff
@@ -46,6 +46,25 @@ def __init__(self, parsed: bool = False) -> None:
     def body(self) -> List[Tuple[Optional[Key], Item]]:
         return self._body
 
+    def unwrap(self, recursive: bool = True) -> str:
+        unwrapped = {}
+        for k, v in self._body:
```
frostming, line 51: We don't need to iterate over the `_body`, `for k, v in self.items()` can retrieve all key and value pairs, without any format control elements.

tomlkit/items.py
```diff
@@ -468,6 +472,8 @@ def __reduce__(self):
     def __reduce_ex__(self, protocol):
         return self.__class__, self._getstate(protocol)
 
+    def 
+
```
frostming, line 476: Need to be cleaned up.

tomlkit/items.py
```diff
@@ -957,6 +984,15 @@ def __init__(self, value: list, trivia: Trivia, multiline: bool = False) -> None
         self._multiline = multiline
         self._reindex()
 
+    def unwrap(self, recursive: bool = True) -> str:
```
frostming, line 987: Are there any cases that we don't want recursive unwrap? `list(array)` should do the trick.

frostming (changes requested):

syntapy:
Ok I think I just need to add tests for
- [ ] `container.Container`
- [ ] `container.OutOfOrderTableProxy`
- [ ] `items.Item`
- [ ] `items.Integer`
- [ ] `items.Float`
- [ ] `items.Bool`
- [ ] `items.DateTime`
- [ ] `items.Date`
- [ ] `items.Time`
- [ ] `items.Array`
- [ ] `items.AbstractTable`
- [ ] `items.String`
- [ ] `items.AoT`
- [ ] `items.Null`

tomlkit/items.py
syntapy: So you mean just returning `list(self)`?

> Are there any cases that we don't want recursive unwrap?

Is that only applicable to the `items.Array` object or do you want me to make all of of the `unwrap` methods recursive?

syntapy: Ok, I think I see something of what you're saying. 

So if a create an `Array` using `array = items.Array([item(2), item(2.5), item(False)], trivia])`, the third one (boolean) gets turned into a ppo but not the others

This makes the recursive `unwrap` fail on that

But if I do `return list(self)` then the `Integer` and `Float` items in indices 0 and 1 don't come out as ppo

frostming: I mean `recursive=False` is equal to `list(array)`, so we don't need this optional flag and make all `unwrap` recursive.

tomlkit/container.py
```diff
@@ -46,6 +48,25 @@ def __init__(self, parsed: bool = False) -> None:
     def body(self) -> List[Tuple[Optional[Key], Item]]:
         return self._body
 
+    def unwrap(self) -> str:
+        unwrapped = {}
+        for k, v in self.items():
+            if k is None:
+                continue
+
+            if not type(k) == str:
```
frostming, line 57: `type(k) == str` is not recommended against `isinstance`. For this check, we can replace to `isinstance(Key)`

tomlkit/container.py
```diff
@@ -46,6 +48,25 @@ def __init__(self, parsed: bool = False) -> None:
...
+                k = k.key
+
+            if not type(v) == str:
+                v = v.unwrap()
```
frostming, line 61: Might be better to call `unwrap()` and catch `AttributionError`.

frostming (changes requested):

tomlkit/container.py
syntapy: ok I have done that

now its that because of a possibility that `k` and `v` are tomlkit types  that inherit from `str`?

syntapy: I think I need to do some similar stuff with the tests that use the `type(..)` function. I will let you know when I get that done in the next day or so

tomlkit/container.py
```diff
@@ -46,42 +46,33 @@ def __init__(self, parsed: bool = False) -> None:
...
-                continue
+    def _unwrap_inner(self, k, v, d):
+        if k is None:
+            return
 
-            k = k.key
+        k = k.key
+        v = v.value
+
+        if isinstance(v, Container):
```
frostming: It is safe to use `isinstance(v, Item)` here since the child of a container can't be a container.

syntapy: I tried that, not sure I did it correctly

tomlkit/items.py
```diff
@@ -1675,6 +1728,15 @@ def __init__(
         for table in body:
             self.append(table)
 
+    def unwrap(self) -> str:
+        unwrapped = []
+        for t in self._body:
+            if is_tomlkit(t):
```
frostming, line 1734: Also, a child of an array must be an item. `isinstance(t, Item)` is enough for this case. But this also works

syntapy: Done

frostming (approved):

syntapy:
@frostming Is it OK if I make a PR to bump the version? If so which version? 0.11.0 or 0.10.2?

I am trying to use the version of tomlkit with these changes in a [PR](https://github.com/iterative/dvc/pull/7468)

PR #195, opened 2022-05-24
Remove the is_tomlkit check

frostming:


syntapy:
Yeah, thats definitely more elegant than the way I had it in my first PR

Issue #200, opened 2022-06-18
Regression: inconsistent line endings


radoering:
#149 introduced a regression especially for Windows users. If your toml file contains only windows line endings (`"\r\n"`), you may get mixed line endings when you read, change and write the toml file.

Example:

```
toml_path = str(tmpdir / "pyproject.toml")
with open(toml_path, "wb+") as f:
    f.write(b"a = 1\r\nb = 2\r\n")

f = TOMLFile(toml_path)
content = f.read()
content["c"] = 3
f.write(content)

with open(toml_path, "rb") as f:
    print(f.read())
```

Output: `b'a = 1\r\nb = 2\r\nc = 3\n'` (the last line ending is `"\n"` instead of `"\r\n"`)

PR #201, opened 2022-06-18
Keep consistent line endings

radoering:
Resolves: #200

With this change, when reading a toml file with consistent line endings, changing the content and writing the content into a file, the result is a file with the same consistent line endings as the input file.

Further, default line endings for Windows when creating a new toml file will be Windows line endings (`"\r\n"`) again.

This change is not perfect, because it only ensures that the written file has consistent line endings. Dumping the changed `TomlDocument` as string still has inconsistent line endings. However, keeping consistent line endings for `TomlDocument` seems to require way more effort since hard coded `"\n"` is used in many places.

### DIFF
--- a/tests/test_toml_file.py
+++ b/tests/test_toml_file.py
@@ -62,3 +62,46 @@ def test_mixed_eol(tmpdir):
 
     with open(toml_path, "rb") as f:
         assert f.read() == b"a = 1\r\nrb = 2\n"
+
+
+def test_consistent_eol(tmpdir):
+    toml_path = str(tmpdir / "pyproject.toml")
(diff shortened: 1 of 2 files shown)


frostming (approved):

radoering:
@frostming poetry tests were kind of unintentionally expecting inconsistent line endings... They should be forward compatible now. Please re-run the failing tests.

frostming (approved):
LGTM

frostming:
> Is it possible to add a configuration setting for this?

You are free not to use the `TOMLFile` API and read and parse the document yourself.

PR #212, opened 2022-07-18
Multiple code quality improvements

Secrus:
- Added a few `flake8` plugins and corrected issues that raised
- Added `pyupgrade` (set to 3.6+)
- Added `pycln` linter


.pre-commit-config.yaml
```diff
@@ -19,3 +19,40 @@ repos:
...
+    hooks:
+      - id: pyupgrade
+        args: [--py36-plus]
+
+  - repo: https://github.com/hadialqattan/pycln
+    rev: v2.0.1
+    hooks:
+      - id: pycln
+        args: [--all]
+        exclude: ^.*__init__\.py$ # without this, pycln removes TOMLDocument import
```
frostming, line 52: Add it to `__all__` should fix it

tomlkit/exceptions.py
```diff
@@ -1,5 +1,6 @@
+from __future__ import annotations
```
frostming, line 1: This only works on 3.7+ IIRC?

frostming (changes requested):

tomlkit/exceptions.py
Secrus: Oh, right. I knew I forgot about something. I guess I will need to do some reverting.

.pre-commit-config.yaml
Secrus: Fixed

tomlkit/exceptions.py
Secrus: Reverted.

frostming (approved):
LGTM

PR #221, opened 2022-08-09
bugfix: items.Array __init__ didn't ignore Null that _group_values added.

laixintao:
The issue is, #218 added a `_group_values` cloud add `Null` into the Array values, but the `Array`'s `list.__init__` didn't ignore this value.

Poetry will `copy.copy` an `Array` item (for e.g. here: https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/packages/package.py#L290 ), then there will be an extra `Null` object in the values, so that an Array like this:

```
[tool.poetry]
classifiers = [
# comment
    "a",
    "b",
]
```
originally is `["a", "b"]`, after being `copy.copy`ed, will become `[None, "a", "b"]` due to the `Null`. (Please see the testcase I added)

Then it will cause poetry to fail because there is an extra "None" in classifiers. ( https://github.com/laixintao/iredis/runs/7749027851?check_suite_focus=true )

related: https://github.com/sdispater/tomlkit/pull/218

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -1,5 +1,6 @@
 import math
 import pickle
+import copy
 
 from datetime import date
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


laixintao:
I found this issue in iredis' build:

It coincidentally has a comment line in classifiers:
https://github.com/laixintao/iredis/blob/d0cafb62607868fa9859214d1cefe9ae29d6b273/pyproject.toml#L12

the poetry failure: https://github.com/laixintao/iredis/runs/7749027851?check_suite_focus=true

cc @frostming

frostming:
Thanks for catching this

PR #226, opened 2022-08-11
Fix memory leak by reverting to per-method cache decorator

Secrus:


frostming:
Closed in favor of #227

PR #235, opened 2022-09-13
fix: fix a bug that escape chars are lost after concat with another string

### LINKED ISSUE #231: Escape characters lost after concatenating string values
When working with AWS samconfig.toml, you have to specify parameter overrides as literal strings:
```
[default.deploy.parameters]
parameter_overrides = "<parameter_name>=\"<parameter_value>\""
```
When loading this `.toml` with `tomlkit`, the backslashes are lost in translation. So a roundtrip: `tomlkit.dump(tomlkit.load(config))` results in:
```
[default.deploy.parameters]
parameter_overrides = "<parameter_name>="<parameter_value>""
```
And this breaks the `parameter_overrides` string, since the intermediate `"` now break apart the string.

How is it possible to load strings as literal strings using `tomlkit`?

frostming:
Fix #231

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,12 +27,12 @@ jobs:
         python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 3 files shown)


aGitForEveryone:
Thanks!

Issue #161, opened 2021-12-23
Add a doc site


frostming:

frostming:
Hi, @sdispater , any chance adding the github integration of RTFD? The toolkit project on RTFD is owned by me now, I can add you as maintainer.

remram44:
The link in the README is 404. What do you own @frostming?

webknjaz:
> [readthedocs.org/projects/tomlkit/builds](https://readthedocs.org/projects/tomlkit/builds/) is failing for some reason

Here's the fix: https://github.com/sdispater/tomlkit/pull/251.

PR #238, opened 2022-09-16
fix: sphinx build failed

wsgalaxy:
Sphinx build failed with:
> Theme error:
> no theme named 'furo' found (missing theme.conf?)

Use builtin "classic" theme.


wsgalaxy:
This may be the reason why readthedocs failed to build the documents, see https://github.com/sdispater/tomlkit/issues/161 .

frostming:
How did you build the docs, `furo` is specified in both `docs/requirements.txt` and `pyproject.toml`, this PR doesn't seem correct to me.

webknjaz:
@wsgalaxy for RTD, there's a need to create a `readthedocs.yaml` config, pointing it at the required dependencies and Python env.

Issue #237, opened 2022-09-15
Python 3.11 str enum inconsistency


Tinche:
Hi folks,

not sure if you support 3.11 yet (or even plan on it), but I'm working to add support for it to cattrs. Since cattrs has a preconfigured converter for tomlkit, the test suite found a discrepancy.

```python
from enum import Enum

from tomlkit import dumps


class Test(str, Enum):
    A = "A"


print(dumps({Test.A: "a"}))
```

On 3.10:

```
A = "a"
```

On 3.11:

```
Test.A = "a"
```

frostming:
The enum interface is a big inconsistency in the historic versions of Python. And Python 3.11 tries to address it.

Given this script:
```python
import tomlkit
from enum import Enum, IntEnum

class Foo(str, Enum):
    A = "A"

class IntFoo(IntEnum):
    ONE = 1

print("str(Enum):", str(Foo.A), "f-string:", f"{Foo.A}")
print("str(IntEnum):", str(IntFoo.ONE), "f-string:", f"{IntFoo.ONE}")
print("str key:", tomlkit.key(Foo.A).as_string())
print("str value:", tomlkit.item(Foo.A).as_string())
print("int value:", tomlkit.item(IntFoo.ONE).as_string())
```
Python 3.10 output, note that how f-string output differs from str().
```
str(Enum): Foo.A f-string: A
str(IntEnum): IntFoo.ONE f-string: 1
str key: A
str value: "A"
int value: IntFoo.ONE
```
Python 3.11 output:
```
str(Enum): Foo.A f-string: Foo.A
str(IntEnum): 1 f-string: 1
str key: Foo.A
str value: "A"
int value: 1
```

we wouldn't like to address this kind of mess in tomlkit, users should guarantee to pass primitive values as keys and values.

PR #239, opened 2022-09-17
Add Python 3.11 release candidate 2 to the testing

cclauss:
Python 3.11 will be on average 22% faster than Python 3.10 so let’s give it a spin... https://www.python.org/download/pre-releases

Related to #237

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11-dev]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11.0-rc - 3.11"]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install Poetry
         shell: bash
         run: |
-          curl -fsSL https://install.python-poetry.org | python - -y --version 1.1.15
+          curl -fsSL https://install.python-poetry.org | python - -y --version 1.2.1
       - name: Update PATH
         if: ${{ matrix.os != 'Windows' }}
         run: echo "$HOME/.local/bin" >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -24,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11-dev]
```
frostming, line 29:
```suggestion
        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11.0-beta - 3.11.0"]
```
To include pre-releases only

cclauss: `python-version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11.0-rc - 3.11"]`

frostming: That is fine. Thanks

cclauss:
> Successfully set up CPython (3.11.0-rc.2)

cclauss:
My sense is that supporting Python 3.11 using Poetry will require dropping Python 3.6 support.
* https://github.com/python-poetry/poetry/releases

frostming:
> My sense is that supporting Python 3.11 using Poetry will require dropping Python 3.6 support.

True, that is why it was `1.1.5` before. Since `tomlkit` is a lower-level library, I prefer it to stay supporting 3.6 a bit longer. Dropping the support only because the testing infra isn't happy doesn't make sense to me.

cclauss:
Poetry is not test infra.  Poetry is how the tomlkit is built.
* https://devguide.python.org/versions
* Python 3.11 is ~22% faster than 3.10 so there will be a lot of interest in migrating when it is released next week.
* Please let this test run so we can ensure that it passes and then I will make mods so we can get 3.6 and 3.11 to pass.

.github/workflows/tests.yml
```diff
@@ -46,6 +49,8 @@ jobs:
         shell: bash
```
frostming, line 49: This step shouldn't always be executed

cclauss: I tried `if: ! startsWith()` which seems to invalidate the entire file.  Removing the space between `!` and `startsWith()` seems to be the solution.

frostming:
LGTM!

Issue #214, opened 2022-07-27
float addition with tomlkit.items.Integer raises a NotImplementedType Exception


Poofjunior:
Title says it all. Adding a tomlkit.items.Integer with a float will raise a NotImplementedType exception.

Here's the breaking code:
````
content = '[table]\nmy_int = 2048'
doc  = parse(content)

doc['table']['my_int'] + 5.0
````

## System Info
Python version: 3.8.13
Tomlkit version: 0.11.1
OS: Windows 10

## Other stuff
Some quick googling looks like the same issue shows up in another project [here](https://github.com/optuna/optuna/issues/3652).

SamirPS:
Do you have the logs of the error ?

Poofjunior:
I do! Here's the stack trace:
````
Traceback (most recent call last):
  File "C:\Users\joshua.vasquez\.conda\envs\spim\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\Users\joshua.vasquez\.conda\envs\spim\lib\site-packages\tomlkit\items.py", line 635, in __add__
    return self._new(result)
  File "C:\Users\joshua.vasquez\.conda\envs\spim\lib\site-packages\tomlkit\items.py", line 665, in _new
    return Integer(result, self._trivia, raw)
  File "C:\Users\joshua.vasquez\.conda\envs\spim\lib\site-packages\tomlkit\items.py", line 606, in __new__
    return super().__new__(cls, value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NotImplementedType'
````

SamirPS:
I didn't check the code but if you convert the doc['table']['my_int'] to int it's works ?

SamirPS:
i thinks it's line 640 on items.py :         result = super().__add__(other)

SamirPS:
if you do : doc['table']['my_int'] + 5 it's work

PR #271, opened 2023-02-06
Fixed float subtraction with tomlkit.items.Integer

### LINKED ISSUE #270: Float subtraction with tomlkit.items.Integer raises a NotImplementedType exception
Subtraction with an `tomlkit.items.Integer` and a basic python `float` leads to a `NotImplementedType` error.

The issue seems to be the same category like the addition issue #214 (with the [same solution](https://github.com/sdispater/tomlkit/pull/215/files)?).

stack trace:
```console
>>> doc['table']['my_int'] - 5.0
  File "<stdin>", line 1, in <module>
  File "/Users/robot/Library/Python/3.10/lib/python/site-packages/tomlkit/items.py", line 657, in __sub__
    return self._new(result)
  File "/Users/robot/Library/Python/3.10/lib/python/site-packages/tomlkit/items.py", line 673, in _new
    return Integer(result, self._trivia, raw)
  File "/Users/robot/Library/Python/3.10/lib/python/site-packages/tomlkit/items.py", line 617, in __new__
    return super().__new__(cls, value)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NotImplementedType'
```

minimal code which reproduces the error:

```py
import tomlkit
content = '[table]\nmy_int = 2048'
doc = tomlkit.parse(content)
doc['table']['my_int'] - 5.0
```

used tomlkit version: `0.11.6`

zerocewl:
closes #270

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -585,6 +585,20 @@ def test_add_sum_int_with_float():
     assert doc["table"]["my_int"] == 2053.3
 
 
+def test_sub_float_from_int():
+    content = "[table]\nmy_int = 2048"
+    doc = parse(content)
+    doc["table"]["my_int"] -= 5.0
+    assert doc["table"]["my_int"] == 2043.0
+
+
+def test_sub_int_from_float():
+    content = "[table]\nmy_int = 2048.0"
+    doc = parse(content)
+    doc["table"]["my_int"] -= 5
+    assert doc["table"]["my_int"] == 2043.0
+
+
 def test_integers_behave_like_ints():
     i = item(34)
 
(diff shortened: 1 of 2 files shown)


frostming:
This fix doesn't look good to me, we don't expect an `Integer` as the result of substracting a float.

zerocewl:
@frostming 

have you looked at the already implemented definition of `def __add__(self, other)` [here](https://github.com/sdispater/tomlkit/blob/be9af5549a765442b073e29f51083463c371e6ad/tomlkit/items.py#L649) which i mentioned in #270?

```py
    def __add__(self, other):
        return self._new(int(self._raw) + other)
```

i think it is the same and worked to solve #214.

Maybe you can explain why this won't fit here?

frostming:
Superseded by #277

PR #280, opened 2023-04-12
Feat: Support `Pathlib.Path` by casting to `str`

caniko:


frostming:
That's beyond what a toml parser should do. It's trivial to convert a Path to a string.

caniko:
> That's beyond what a toml parser should do. It's trivial to convert a Path to a string.

What about date and time objects? They are supported, they are also trivial to convert.

frostming:
> > That's beyond what a toml parser should do. It's trivial to convert a Path to a string.
> 
> What about date and time objects? They are supported, they are also trivial to convert.

The example isn't appropriate. Datetimes are [native objects in TOML](https://toml.io/en/v1.0.0#offset-date-time) and in Python they are `datetime` objects. `pathlib.Path` isn't, in the TOML spec there isn't such type and you are just requesting to "borrow" string to represent it. What's more, it can't be converted back from a TOML string.

I could have not replied this, but you seem to misunderstand.

caniko:
I understand, I think this strictness hurts the utility of the package.

I'll make a package downstream that don't have these shackles.

frostming:
> Both json and PyYaml parse Path just fine.

How did you come to this conclusion?
```python
>>> from pathlib import PurePath
>>> doc = {"path": PurePath("where")}
>>> json.dumps(doc)
TypeError: Object of type PurePosixPath is not JSON serializable
```

```python
>>> yaml.safe_dump(doc)
RepresenterError: ('cannot represent an object', PurePosixPath('where'))
```

Also, it isn't a process of "parse", it's serialization(object -> representation).

caniko:
@frostming what you are doing right now is called paltering; you are leaving out key facts to make your own narrative.

You are using the `safe_dump`, method, which is strict. The standard `dump` method should dump it just fine.

frostming:
> You are using the `safe_dump`, method, which is strict. The standard `dump` method should dump it just fine.

True, here is the output of `dump`:

```python
>>> print(yaml.dump(doc))
path: !!python/object/apply:pathlib.PurePosixPath
- where
```

I didn't mention this due to the following reasons

- YAML is a rather flexible format, more than JSON and TOML, and I don't think this feature can generalize to the other two formats. Something is supported on YAML doesn't mean it is the common practice
- Most practices recommends `safe_dump` over `dump`, dumping an object like that isn't a good idea.

Anyway, the other examples I provided strongly support my point of view. I don't understand why you have to use such a serious word as "deception".

PR #282, opened 2023-04-19
Add types to the `append()` methods of the `Table` and `InlineTable` classes

### LINKED ISSUE #281: Missing types in the `append()` method of the `Table` class
Hi! 👋 

I think the types for the [`append()` method of the `Table` class](https://github.com/sdispater/tomlkit/blob/master/tomlkit/items.py#L1577) are missing, as this method is part of the "public API". If adding them makes sense and I can help, let me know. 

Thanks! 😄

joaopalmeiro:
Closes #281

### DIFF
--- a/tomlkit/items.py
+++ b/tomlkit/items.py
@@ -1574,7 +1574,7 @@ class Table(AbstractTable):
             self.display_name,
         )
 
-    def append(self, key: Union[Key, str], _item: Any) -> "Table":
+    def append(self, key: Union[Key, str, None], _item: Any) -> "Table":
         """
         Appends a (key, item) to the table.
         """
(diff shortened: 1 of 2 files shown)


frostming:
Do you mind helping fix the lint error?

joaopalmeiro:
> Do you mind helping fix the lint error?

Sure!

tomlkit/items.py
```diff
@@ -1574,7 +1574,7 @@ def __copy__(self) -> "Table":
             self.display_name,
         )
 
-    def append(self, key, _item):
+    def append(self, key: Union[Key, str], _item: Any) -> "Table":
```
frostming, line 1577: It seems key can be `None`, just add `None` to the Union

joaopalmeiro: Sure! So, these methods will have the same types as the `append()` method of the `Container` class, right?

joaopalmeiro: @frostming , added `None` to the `append()` and `raw_append()` methods and fixed the linting errors. Can you check the MR again, please?

frostming (approved):

PR #287, opened 2023-05-02
Update README.md

chapnitsky:
A small addition by specifing that poetry should be installed by pip before using poetry install.


Secrus:
Poetry should not be installed by `pip`, but rather using `pipx` or Poetry installation script. In any way, this is already mentioned above `If you are using Poetry...`, with a link to the Poetry website. This addition is worthless for `tomlkit` and just wrong. Please close.

Issue #290, opened 2023-05-12
Add indent control for multi-line `Array`


goodboy:
More or less a control for an `indent: str` when `Array._multiline == True`.

The `Trivia.indent` doesn't seem be the right place because it's in reference to the parent container?
Also, **only when** it's a multiline array does this indent *exist*; I'm thinking add an `Array._multiline_indent: str` ?

Issue #289, opened 2023-05-12
Support `bidict` inputs?


goodboy:
Similar to #288 but instead supporting the pretty popular [`bidict`](https://github.com/jab/bidict) lib 😎 

Seems like (again) it should be simple to [match against a `collections.abc.MutableMapping` instead of a `dict`](https://github.com/sdispater/tomlkit/blob/master/tomlkit/items.py#L151)?

I'm probably going to put up a patch for both this and #288.

Issue #288, opened 2023-05-12
Support `Decimal` serialization?


goodboy:
Love this lib btw, it's been a huge headache trying to write a certain style of `TOML` with anything else; code base is also very easy to nav, so big 👍🏼 

So the [`tomli` lib has this](https://github.com/hukkin/tomli#construct-decimaldecimals-from-toml-floats) and it'd sure be handy to have here instead of having to traverse a tree to convert everything to `str` first 😂 

I think to start it'd be good enough to just serialize to `String` since presumably on parse the client code should know if it wants to cast to decimal (and obvsly TOML doesn't spec a decimal type); though i guess it wouldn't be too hard to add some logic + flag to `.parser.Parser._parse_number()` ?

goodboy:
As per discussion in #294 it makes more sense to just expect users to define a custom encoder since there's no type support in the TOML spec for `Decimal`:
https://github.com/sdispater/tomlkit/pull/294#issuecomment-1598233886

Those looking for a solution can defer to @frostming's comment linked above.

PR #294, opened 2023-05-25
Support `Decimal` and `MutableMapping` codecs, add a multiline-`Array` indent control

goodboy:
In an attempt to resolve the feature request issues i submitted:
- #288
- #289
- #290

---
#### Summary:
- add a `Array._multiline_indent: str` and expose it through a `indent: str` kwarg to `.multiline()`.
  - main purpose is to be able to *style* the indent level for user preferred indentation of say a multiline array of tables B)
- add a branch to the `.items.item()` encoder to handle `decimal.Decimal` and match against `MutableMapping` instead of `dict`.
- tests for all of it!

---
#### Possibly still todo?
- [ ] Do i need to submit more for `Decimal` and `MutableMapping` changes?
  - currently i added 2 new tests for decimal encoding in the best places i could find.
  - don't know if mut map tweaks require anything extra other then maybe some tests maybe using a custom type? ( i feel like that is just testing the protocol and thus stdlib stuff 😂 )
- [ ] tests for the inline `Array` indent parameter:
  - probably in [tests_items.py](https://github.com/sdispater/tomlkit/blob/master/tests/test_items.py#L322) i presume?

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -187,6 +187,12 @@ def test_integer():
     assert isinstance(i, Integer)
 
 
+def test_decimal():
+    d = tomlkit.decimal("34.56")
+
+    assert isinstance(d, String)
+
+
 def test_float():
     i = tomlkit.float_("34.56")
 
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -2,6 +2,7 @@ import copy
 import math
 import pickle
 
+from decimal import Decimal
 from datetime import date
 from datetime import datetime
 from datetime import time
@@ -88,6 +89,16 @@ def test_item_base_has_no_unwrap():
 
 def test_integer_unwrap():
     elementary_test(item(666), int)
+
+
+def test_decimal_unwrap():
+    """Ensure a decimal unwraps as a string
+    after TOML encode.
+    """
+    elementary_test(
+        item(Decimal("0.001")),
+        str,
+    )
 
 
 def test_float_unwrap():


goodboy:
Bah, going through rebase to master 🏄🏼

goodboy:
Also, if this patch is likely to never be accepted or needs to be changed to be accepted please let me know asap since we'd much rather not have to maintain our own fork :)

frostming:
a better way to support this and similar cases is to allow passing a custom decoder to item() function.

goodboy:
@frostming ah ok, I'm fine to change it to do that.

(also sorry i missed your comment don't have my email hooked up rn for GH 😂)

can you show an example that might already exist for doing this btw in the code base?

frostming:
You can convert a decimal to a float representation in TOML, but can't convert it back, and Decimal isn't listed as the data types defined by [TOML spec](https://toml.io/en/index), so it's not worth to expose a new api and item type for it.

Instead, I'd propose something like this:

```python
import tomlkit
a = Decimal('34.56')

def my_encoder(o):
    if isinstance(o, Decimal):
        return tomlkit.float_(o)
    raise ValueError(f"Invalid type {type(value)}")

item = tomlkit.item(a, encoder=my_encoder)   # a Float object
v = {"a": Decimal("34.56")}
print(tomlkit.dumps(v, encoder=my_encoder)
# a = 34.56
```

tomlkit/items.py
```diff
@@ -1154,7 +1179,12 @@ def _iter_items(self) -> Iterator[Item]:
         for v in self._value:
             yield from v
 
-    def multiline(self, multiline: bool) -> Array:
+    def multiline(
+        self,
+        multiline: bool,
+        indent: str = " "*4,
```
frostming, line 1185: Prefer to use int indent, to prevent passing arbitrary strings.

goodboy:
> so it's not worth to expose a new api and item type for it.
> In this way users can convert custom types, not only Decimal, to TOML items in tomlkit. And it is clearer that users can't convert it back(parse anything as Decimal)

@frostming works for me. So I guess i'll just drop all the `Decimal` related changes then and rebase this branch a bit.

Any other quiffs and/or do you have any suggestions for the tests I still need to do?

frostming:
@goodboy yes, I've proposed a change in #296

goodboy:
@frostming thanks for the follow up!

i will hopefully get back to this PR this week. Will check out that landed PR and change our stuff to match first, then come back here and drop the unneeded history.

Still kinda looking for a little guidance on how thorough the new tests for the multi-line arrays should be?
i have some basic ideas but don't want to push too deep if y'all aren't that concerned about it.

Issue #305, opened 2023-07-26
Missing setup.py on releases?


luizribeiro:
While packaging `tomlkit` 0.11.8 for NixOS we noticed that starting on 0.11.7 the released packages no longer include `setup.py` (see full discussion on https://github.com/NixOS/nixpkgs/pull/245440).

While `setup.py` is no longer actually necessary, this creates a circular dependency to build `tomlkit` moving forward - as `tomlkit` itself would require `poetry` to be able to build `tomlkit`, while `poetry` itself depends on `tomlkit` to be built.

Was the removal of `setup.py` intended?

luizribeiro:
Ah, it seems like this happened because of https://github.com/python-poetry/poetry-core/pull/318

I guess we could explicitly set `tool.poetry.build.generate-setup-file` here so we can still build `tomlkit` without `poetry`?

I'm going to put up a PR for that.

frostming:
> this creates a circular dependency to build `tomlkit` moving forward - as `tomlkit` itself would require `poetry` to be able to build `tomlkit`, while `poetry` itself depends on `tomlkit` to be built.

This is not true, `tomlkit` only requires `poetry-core` to build and the latter has zero-dependency, it doesn't vendor tomlkit either. You may need to investigate carefully to find out the real cause.

luizribeiro:
Thank you for the clarification, @frostming!

It looks like nixpkgs has the wrong dependencies for `poetry-core` indeed.

Feel free to close this issue (and the PR) if you feel like there shouldn’t be any changes upstream.

PR #306, opened 2023-07-26
fix(#305): generate setup.py file

luizribeiro:
Since poetry 1.3.0 setup.py files are no longer generated by default. See https://python-poetry.org/blog/announcing-poetry-1.3.0/#generate-setup-file--false

This causes an issue where in order to build `tomlkit` we need to use `poetry` itself.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,9 @@ include = [
   { path = "CHANGELOG.md", format = "sdist" },
 ]
 
+[tool.poetry.build]
+generate-setup-file = true
+
 [tool.poetry.dependencies]
 python = ">=3.7"
 


luizribeiro:
I'm actually not 100% sure we should merge this in, but putting the PR up anyway for discussion.

The blog post linked on the PR mentions that this setting will be removed too eventually (though it doesn't hint at when that will happen), so this feels like an intermediate solution for distributions which require `poetry` to build `pyproject.toml`-based packages (which leads to the circular dependency mentioned on #305).

Thoughts?

luizribeiro:
Alternatively we could switch the build system for `tomlkit` from using `poetry` to use `setuptools` instead with:

```toml
[build-system]
requires = ["setuptools>=40.8.0", "wheel"]
build-backend = "setuptools.build_meta:__legacy__"
```

But I'm not sure what other implications that would have.

Secrus:
This is completely wrong. You don't need `poetry` to build `tomlkit`. You can use [`build`](https://pypi.org/project/build) and it will build `tomlkit` just fine without the need for (now considered deprecated) `setup.py` script.

luizribeiro:
Yeah it was also clarified that you can build with `poetry-core`, which doesn't depend on `tomlkit`. Thanks @Secrus

Issue #327, opened 2024-01-05
Error when removing the first element from array: AttributeError: 'NoneType' object has no attribute 's'


pinal005:
seeing this edge case while removing the first element from the array:

Below code works fine if there is a space before the second element(i.e. "green" in below example) and try to remove the first element.
```
import tomlkit
doc = tomlkit.loads("""
[colors]
primary = ["red", "green","blue"]
""")

doc["colors"]["primary"].remove("red")
```

But errors out if there is no space before the second element "green"
```
import tomlkit
doc = tomlkit.loads("""
[colors]
primary = ["red","green","blue"]
""")
doc["colors"]["primary"].remove("red")
```

Error:
```
tomlkit/items.py", line 1415, in __delitem__
    and "\n" not in self._value[idx].indent.s
AttributeError: 'NoneType' object has no attribute 's'
```

PR #328, opened 2024-01-05
indent exist check before accessing its property

pinal005:
fixes https://github.com/sdispater/tomlkit/issues/327


frostming (approved):

aerickson:
Currently experiencing this issue. 

Can we get a release cut that includes this? Thanks!

PR #345, opened 2024-05-07
fix: don't add sign if the float is negative

### LINKED ISSUE #341: Sign change when doing math gives invalid number
Bug:
```
>>> d = tomlkit.parse('x=-3.1415')
>>> d['x'] *= -1
>>> print(tomlkit.dumps(d))
```
returns:
```
a=+3.1415

>>> d['x'] *= -1
>>> print(tomlkit.dumps(d))
```
returns:
```
a=--3.1415
```

Subsequent parsing raises `tomlkit.exceptions.InvalidNumberError`.

Of course there is an easy workaround using type conversion: `d['x'] = float(d['x'])*-1`.

frostming:
Fix #341

Signed-off-by: Frost Ming <me@frostming.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [unreleased]
+
+### Fixed
+
+- Remove the extra minus sign added to the float value after calculation. ([#341](https://github.com/python-poetry/tomlkit/issues/341))
+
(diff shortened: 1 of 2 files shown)


JamesParrott:
This is great.  Thanks @frostming .  There's a similar issue with Integers too.

PR #356, opened 2024-05-31
fix: mypy errors when using tomlkit to modify toml document

### LINKED ISSUE #326: MyPy errors when using tomlkit to modify toml documents
When using tomlkit to modify a tomldocument, mypy throws a bunch of errors that I don't know how to fix.
Example:
```python
import tomlkit
from pathlib import Path

path = Path("data.toml")

path.write_text("""
[foo]
bar = 'baz'
[foo.bla]
blub = 'blib'
""")

data = tomlkit.parse(path.read_text(encoding="utf-8"))
data["foo"]["bar"] = "quux"
data["foo"]["bla"]["blub"] = "bleb"
path.write_bytes(tomlkit.dumps(data).encode("utf-8"))
```
Output
```python
tomlkit_issue.py:14: error: Unsupported target for indexed assignment ("Item | Container")  [index]
tomlkit_issue.py:15: error: Value of type "Item | Container" is not indexable  [index]
tomlkit_issue.py:15: error: Unsupported target for indexed assignment ("Any | Item | Container")  [index]
Found 3 errors in 1 file (checked 1 source file)
```

Is there any fix for this?
Any help is greatly appreciated :)

waketzheng:
Fixes #326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Fix mypy errors when using tomlkit to modify toml document. ([#326](https://github.com/python-poetry/tomlkit/issues/326))
 - Fix the incompatiblity with 3.13 because of the `datetime.replace()` change. ([#333](https://github.com/python-poetry/tomlkit/issues/333))
 - Revert the change of parsing out-of-order tables. ([#347](https://github.com/python-poetry/tomlkit/issues/347))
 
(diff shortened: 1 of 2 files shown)


waketzheng:
Mini demo to check result:
```py
from pathlib import Path
import tomlkit
 
doc = tomlkit.parse(Path('pyproject.toml').read_text())
poetry_conf = doc['tool']['poetry']
print(poetry_conf)
poetry_conf['version'] = '0'
```

frostming:
But that is not right, you can't assume an abitrary item is a table.

You can't make that assumption in vanilla python either:

```python
doc: dict[str, Any] = get_data()
print(doc['tool']['abc'])  # what type is doc['tool']?
```

PR #357, opened 2024-06-04
Update tests action

waketzheng:
1. Update cache action version
2. Remove quotes around 3.11
3. Use map to replace list comprehension

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-13, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", 3.12, 3.13]
+        python-version: [3.7, 3.8, 3.9, "3.10", 3.11, 3.12, 3.13]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         id: full-python-version
         shell: bash
         run: |
-          echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
```
frostming, line 37: Can you change this to the new `GITHUB_OUTPUT` syntax?

waketzheng: Do you mean: `echo "{name}={value}" >> "$GITHUB_OUTPUT"`?

frostming: Right

.github/workflows/tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
...
+          echo name=version::$(python -c "import sys; print('-'.join(map(str, sys.version_info)))") >> "$GITHUB_OUTPUT"
```
frostming, line 37:
```suggestion
          echo version=$(python -c "import sys; print('-'.join(map(str, sys.version_info)))") >> "$GITHUB_OUTPUT"
```

frostming (approved):

Issue #381, opened 2024-09-20
Missing newline at end of file with multiple array of tables can corrupt during dumping


gdiepen:
When creating a pipeline using pipenv, I ran into a weird situation where the Pipfile would get corrupted in certain cases, when there would be multiple entries in an array of tables and the last one was at the end of the file AND that does not have a final newline at the end

The following python code reproduces the problem with tomlkit 0.13.2

```python
import tomlkit

toml_str = '''[[products]]
name = "Hammer"

[foo]

[bar]

[[products]]
name = "Nail"'''

parsed = tomlkit.loads(toml_str)

print(tomlkit.dumps(parsed))
```

Running this python code will give the following (wrong) output:
```
[[products]]
name = "Hammer"

[[products]]
name = "Nail"[foo]

[bar]
```

As you can see, the `foo` table entry is put directly after the last entry of products

If you add a trailing newline at the end of the file, this problem does not occur.

PR #382, opened 2024-09-28
Add a final line-ending if multi-line input string

gdiepen:
Without this final line-ending, reordering steps will fail because tables will be added directly to the previous value (see issue #381)

when creating the string from the Container again, we remove any final line-ending if we added it initially to ensure we do not modify the original behavior

This modification only fails on two unit tests. However, I do not know enough about the TOML specifiation to see why this is a problem. Is there anybody that could help out and point me to a direction on how I can fix these final two unit tests?

### DIFF
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -33,11 +33,12 @@ class Container(_CustomDict):
     This class implements the `dict` interface with copy/deepcopy protocol.
     """
 
-    def __init__(self, parsed: bool = False) -> None:
+    def __init__(self, parsed: bool = False, last_line_newline_added: str = "") -> None:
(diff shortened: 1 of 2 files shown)


gdiepen:
Checked the two unit tests that are failing, one is doing a parse, copy, as_string of `[foo]\nbar=1\n` and after the three steps, it is now missing the final newline.

Would this not need to actually have the trailing newline?


The other unit tests that now fails is parsing the value:
`tomlkit.array("""[\n"alpha",\n"omega"\n]""")`

Not sure how I would need to fix this.

An alternative to my current approach of adding a final new-line to the input string would be to fix this when reordering the items in an array instead.

PR #407, opened 2025-03-21
Add `repr()` to items

DimmaDont:
This adds `repr()` to a few items as seen in #406


DimmaDont:
Removed from `Array` and updated a test to catch it.

Issue #431, opened 2025-06-26
Add DottedKey to a super table gives wrong output


deltamarnix:
I am trying to set dotted keys on an already existing super table. This seems to go wrong during the render method, because the code thinks I want to render everything as a super table.

My code:
```py
import tomlkit

doc = tomlkit.loads("""
[output.netcdf_scalar]
value = 5
""")

output_table = doc["output"]
output_table[tomlkit.key(["csv", "path"])] = "output.csv"

print(tomlkit.dumps(doc))
```

Output:
```toml
csv.path = "output.csv"

[output.netcdf_scalar]
value = 5
```

Expected:

```toml
[output.csv]
path = "output.csv"

[output.netcdf_scalar]
value = 5
```

In the case of rendering the toml, it will say that the item "csv" is dotted, while "output" is not dotted, and "csv" is a super table, so it will not render "output" as prefix. But just adding the prefix will be wrong, because you actually need to at the `[output.csv]` table as a head, instead of writing `output.csv.path`.

https://github.com/python-poetry/tomlkit/blob/8c963db32784623f8fc076ad4ef9b6b01ab0d542/tomlkit/container.py#L551-L566

I think that the _handle_dotted_key function is in this case not able to indicate that "csv" should not be super:

https://github.com/python-poetry/tomlkit/blob/8c963db32784623f8fc076ad4ef9b6b01ab0d542/tomlkit/container.py#L123-L126

deltamarnix:
Do you think it would be best to make a more automatic function for `is_super_table()` and remove the whole hardcoded setting of `_is_super_table`?

The problem is that `output` itself is now set as `_is_super_table=True` in the parser, but when adding extra keys to `output`, `_is_super_table` should be reevaluated.

Or would it be better update the renderer? `is_super_table()` for `output` indeed returns True, but there are already some extra checks in the renderer:

https://github.com/python-poetry/tomlkit/blob/8c963db32784623f8fc076ad4ef9b6b01ab0d542/tomlkit/container.py#L513-L533

It would mean that we have to revise the check:

```py
     if not table.is_super_table() or ( 
         any( 
             not isinstance(v, (Table, AoT, Whitespace, Null)) 
             for _, v in table.value.body 
         ) 
         and not key.is_dotted() 
     ): 
```

In this case `table.is_super_table()` returns True, so it continues to the `or` statement. All instances are tables, and the `output` key itself is not dotted. So it will skip rendering the header.
[...]

PR #435, opened 2025-07-06
fix: Add DottedKey to a super table gives wrong output (#431)

deltamarnix:
When a table is already a super table, appending dotted items to it will result in wrong behavior. The header was missing, so the final result was an incorrect representation of the document.

Because super_table is already set and can no longer be re-evaluated, I have added a check to the renderer instead, so that it can determine if the header should be rendered.

I have not gone through different edge cases, but it seems that all other tests are still passing.

### DIFF
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -1293,3 +1293,24 @@ name = "Nail"
         "foo": {},
         "bar": {},
     }
+
+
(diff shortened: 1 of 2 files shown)


frostming (approved):

frostming:
Thank you for that

PR #468, opened 2026-04-03
fix: Fix uncontrolled recursion DoS in parser.py

frostming:
Adds a configurable maximum nesting depth (default 100) to prevent RecursionError when parsing deeply nested arrays or inline tables. When the limit is exceeded, a ParseError is raised instead of allowing unbounded recursion that crashes the process.

Fixes #459


spartan8806:
Hey sorry I just now seen It failed 2 checks, I will try and look in to it and get a fix sometime soon.

frostming:
@spartan8806 Are you still working on this? I will close this and you can open the PR yourself.

Issue #483, opened 2026-05-29
Proposal: ~2.5–3× faster parsing, conformance-preserving


tfoutrein:
## TL;DR

I have a set of parser-only optimizations that make `tomlkit.parse()` roughly
**2.4–2.9× faster** (content-dependent) with **no behavioural change**: the full
test suite passes, including **680/680 of the `toml-test` conformance corpus**,
and `ruff` + `mypy --strict` are clean.

Before opening a large PR, I'd like to check your appetite and preferred shape —
the work is naturally splittable into independent, separately-reviewable pieces.

## Why this matters / why I'm being careful

tomlkit is a hot dependency of Poetry (and poetry-core), so I treated
*correctness as non-negotiable*: every change is gated on the conformance suite,
not just round-trip of valid files. (In fact, while validating I caught a subtle
EOF-sentinel collision a naive char-interning optimization can introduce — the
fix is included and is itself covered by `toml-test invalid/control/*-null`.)

## Reproducing the numbers

A dependency-free benchmark is attached (`bench_standalone.py`): run it on
`master`, then on the branch, and compare medians. Measured here (macOS,
CPython 3.11):

```
# master (8694e4d)
$ python bench_standalone.py
  median :  791.6 ms   (1583 us/parse)

# branch
$ python bench_standalone.py
  median :  316.8 ms   ( 634 us/parse)   -> ~2.5x faster
```

Cross-checked with a drift-immune A/B (both versions timed, interleaved, in one
process) on other workloads:

| Workload | master | branch | factor |
|---|---|---|---|
| embedded sample (`bench_standalone.py`) | 792 ms | 317 ms | **×2.50** |
| number-heavy ~3 KB doc | ~2970 ms | ~1045 ms | **×2.85** |
| 262 distinct `toml-test/valid` files | 1747 ms | 717 ms | **×2.44** |

## What the changes do (grouped, so you can pick)

The core idea: the parser advanced **one character per call** (`inc()` ~2.2 M
times on the benchmark). The big wins replace those with **bulk scans** over the
underlying string. Grouped by theme, independently reviewable:

| # | Theme | Idea | Risk |
|---|---|---|---|
| A | **Bulk char scanning** | index-based `Source`; `advance_while`/`advance_until` replace per-char `inc()` loops in bare-keys, whitespace, numbers, comments, single-line strings | medium — hot path, but pure scan-equivalence |
[...]

PR #487, opened 2026-06-03
Add native __contains__ to Container, Table and InlineTable

tfoutrein:
## Summary

First of the small, self-contained, conformance-preserving performance changes discussed in #483 — now also covering `Table` / `InlineTable` per @frostming's review question.

Both `Container` and the table types (`Table`, `InlineTable`, via `AbstractTable`) inherit `__contains__` from `MutableMapping`, so `key in x` goes through `self[key]` → `__getitem__` → `item()`, which resolves and returns the value — and constructs a `NonExistentKey` exception on every *absent* key — only to discard it.

**`Container`** gets a native `__contains__` that resolves the key exactly as `item()` does (`str` → `SingleKey`; a non-`str`/non-`Key` argument still raises `TypeError`) and probes `self._map` directly. For an out-of-order table (a tuple index) it still builds the `OutOfOrderTableProxy`, so its validation runs exactly as before. This also speeds up `Container.__setitem__`, which does `if key in self` on every assignment.

**`Table` / `InlineTable`** get a native `AbstractTable.__contains__` that delegates to the inner container (`return key in self._value`). The inherited mixin never reached the native `Container.__contains__` (it went through `__getitem__`), so the table types didn't benefit transitively; delegating fixes that while still building the `OutOfOrderTableProxy` for an out-of-order entry, so validation is unchanged.

## Behaviour — no change

Beyond the existing suite, I ran a differential `in` check (present / absent / out-of-order / dotted-string / `Key`-vs-`str` / non-`str` keys) comparing this branch against `master`, for both `Container` and the table types: identical results, including the `TypeError` on a non-string key and the out-of-order proxy path. A regression test on an out-of-order super-table is included.

## Benchmark

Drift-immune A/B in a single process (native `__contains__` vs the inherited `__getitem__`-based path, interleaved so thermal drift cancels), CPython 3.11:

```
Container (mixed present + absent keys):
  inherited via __getitem__: ~610 ms
  native __contains__:       ~520 ms     → ~1.17x faster

Table / InlineTable membership (key in table):
  present key: ~1.2x faster
  absent key:  ~1.8x faster   (the miss path is where the discarded
                               NonExistentKey construction dominates)
```

## Verification

- `pytest tests/` (incl. the `toml-test` conformance submodule): **972 passed** (969 on `master` + 3 new membership tests), no regressions.
- `ruff check` + `ruff format --check`: clean.
- `mypy` (1.19.1): no new errors. The single `# type: ignore[arg-type]` (in `Container.__contains__`) mirrors the existing `SingleKey(key)` call in `item()`.
- The 3 new membership tests were validated by focused mutation testing — every mutant of the new `__contains__` is killed, and an equivalent mutant correctly survives.

---

<sub>Context: this is harness-assisted work — every change is individually benchmarked and gated (full test suite incl. conformance, a differential guard, ruff + mypy) before it can land, and I review and stand behind each PR personally. Happy to slice the rest of #483 into further small PRs — e.g. the same native `__contains__` for `OutOfOrderTableProxy` (a separate base class), or the bulk char-scanning / header-cache work.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-- Speed up membership tests (`key in container`) with a native `Container.__contains__`, avoiding the inherited `MutableMapping` round-trip through `__getitem__` (which resolves the value and builds an exception on every absent key). ([#483](https://github.com/python-poetry/tomlkit/issues/483))
+- Speed up membership tests (`key in ...`) on `Container`, `Table` and `InlineTable` with native `__contains__` implementations, avoiding the inherited `MutableMapping` round-trip through `__getitem__` (which resolves the value and builds an exception on every absent key). ([#483](https://github.com/python-poetry/tomlkit/issues/483))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [unreleased]
 
+### Changed
+
+- Speed up membership tests (`key in container`) with a native `Container.__contains__`, avoiding the inherited `MutableMapping` round-trip through `__getitem__` (which resolves the value and builds an exception on every absent key). ([#483](https://github.com/python-poetry/tomlkit/issues/483))
+
 ### Fixed
 
 - Fix invalid serialization with a duplicated comma when removing a non-edge element from a parsed inline table. ([#486](https://github.com/python-poetry/tomlkit/pull/486))
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -1233,3 +1233,51 @@ def test_array_item_removal_newline_restore_next() -> None:
     doc["x"].remove("1")
     assert doc.as_string() == expected
     parse(doc.as_string())
+
+
+def test_table_membership_matches_inner_container() -> None:
+    doc = parse('[server]\nhost = "localhost"\nport = 8080\nenabled = true\n')
+    table = doc["server"]
+    assert isinstance(table, Table)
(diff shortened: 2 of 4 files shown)


frostming:
LGTM, do you also look at `Table`? does it also need a native impl?

tfoutrein:
Good call — yes. `Table` and `InlineTable` inherit `__contains__` from the same `MutableMapping` mixin, and it didn't benefit from the native `Container.__contains__` transitively: the mixin goes through `self[key]` → `AbstractTable.__getitem__` → `Container.__getitem__`, never `Container.__contains__`.

I've pushed it to this PR — a one-liner on `AbstractTable`, so it covers both `Table` and `InlineTable`:

```python
def __contains__(self, key: object) -> bool:
    return key in self._value
```

Delegating to the inner container picks up the fast `_map` lookup and still builds the `OutOfOrderTableProxy` for an out-of-order entry, so validation (including the `KeyAlreadyPresent` it can raise) is preserved exactly. Iso-functional across present / absent / `Key`-vs-`str` / non-`str` (`TypeError`) / out-of-order / `InlineTable`; full suite + conformance stays green (972), with a regression test on an out-of-order super-table. Membership micro-bench ≈ 1.2x (present) / 1.8x (absent).

`OutOfOrderTableProxy` has the same mixin but is a separate base class, so it isn't covered here — happy to give it the same one-liner in a follow-up if you'd like. I've updated the PR title/description to match.

frostming (approved):

PR #490, opened 2026-06-05
Scan character runs in bulk while parsing

tfoutrein:
> **Stacked on #489** — the first commit here is #489 (index-based `Source`). Best reviewed/merged after #489; GitHub will collapse this to just the bulk-scan commit once #489 lands. Happy to rebase on request.

## What

The parser advances one character at a time through runs of whitespace, bare-key and number characters, paying a `Source.inc()` call (attribute lookups + a `TOMLChar` build + bounds check) for **every** character of the run.

This adds `Source.advance_while(charset)` / `advance_until(stopset)`, which scan the underlying string in a single pass and update the index + current character only once, and uses them for the leading-whitespace, bare-key and number/date runs. They keep the exact value contract of the `while ... and self.inc()` loops they replace (on return `current` is the first character outside / inside the set, or EOF).

## Benchmarks

Parsing speedup across document sizes/shapes (median, interleaved A/B vs `master`, includes #489):

| document | speedup |
|---|---|
| typical mixed (~4 KB) | 1.32× |
| poetry.lock-like (~64 KB) | 1.26× |
| pyproject.toml | 1.17× |
| array-heavy | 1.16× |
| large flat (~90 KB) | 1.15× |

No regression on any shape.

## Tests

Full suite passes (972 tests, incl. the toml-test conformance submodule). No public API or behaviour change — round-trip output is byte-identical to `master` on a varied corpus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Speed up membership tests (`key in ...`) on `Container`, `Table` and `InlineTable` with native `__contains__` implementations, avoiding the inherited `MutableMapping` round-trip through `__getitem__` (which resolves the value and builds an exception on every absent key). ([#483](https://github.com/python-poetry/tomlkit/issues/483))
 - Speed up parsing by making `Source` index-based: it now tracks an integer position over the input string instead of materializing a list of `(index, char)` tuples up front, so construction is O(1) and state save/restore no longer copies an iterator. ([#489](https://github.com/python-poetry/tomlkit/pull/489))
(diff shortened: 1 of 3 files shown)


frostming:
LGTM, please resolve the conflicts.

BTW: we are merging by "squash and merge" so the stacked PR is inconvenient.

tfoutrein:
Rebased onto `master` — conflicts resolved, and it's now a single commit (the #489 changes dropped since they're already merged). No more stacking, so it should squash-merge cleanly. CI is green. Thanks for the review!

frostming (approved):

PR #492, opened 2026-06-06
Remove the internal TOMLChar wrapper

tfoutrein:
> **Stacked on #489, #490 and #491** — the capstone of that series; best reviewed/merged after them.
> This also **supersedes #488** (interning `TOMLChar`): per @dimbleby's suggestion on that PR, removing the wrapper entirely is the better end-state, so I'd close #488 in favour of this.

## What

After the bulk run-scans (#490/#491), the parser only constructs a `TOMLChar` (a `str` subclass) at run boundaries and uses a handful of its `is_*()` helpers. This removes the class entirely:

- `Source` yields plain `str` characters; `inc()` / `advance_*` read `self[i]` directly.
- End-of-input is detected **positionally** (`_idx >= len` / `Source.end()`) instead of an identity sentinel.
- The remaining character-class checks use module-level frozensets.

A real NUL byte is still rejected as an invalid control char and is never mistaken for end-of-input, since EOF is now positional rather than a value/identity comparison.

## Benchmarks

Median, interleaved A/B vs `master` (includes #489–#491):

| document | speedup |
|---|---|
| large flat, single-line strings (~90 KB) | 5.8× |
| poetry.lock-like (~64 KB) | 2.4× |
| pyproject.toml | 1.9× |
| typical mixed (~4 KB) | 1.6× |

The removal itself adds ~1.1–1.18× over #491. No regression on any shape.

## Tests

Full suite passes (972, incl. the toml-test conformance submodule). On top of that, an **11.5k-input adversarial differential** — EOF/truncation at every prefix length, real-NUL placement in every position, empty/whitespace/BOM, and structural fuzz — is **byte-identical** in output and exception type to `master`. No public API change (`TOMLChar` was not exported).


tfoutrein:
Rebased onto `master` now that #491 has merged — single commit on top of current `master`, no conflicts (composes cleanly with the recent escape/bare-key fixes #493/#497/#501). CI is green. This removes `TOMLChar` entirely, so it supersedes #488 (interning) — I'll close #488 once this lands. Ready for squash-merge. Thanks!

tomlkit/parser.py
```diff
@@ -57,11 +56,22 @@
 CTRL_CHAR_LIMIT = 0x1F
 CHR_DEL = 0x7F
 
+# TOML character classes (formerly the `TOMLChar` constants). The parser works on
+# plain 1-char `str`s read from `Source`; membership tests use these frozensets.
+_SPACES = " \t"
+_NL = "\n\r"
+_WS = _SPACES + _NL
+_BARE = string.ascii_letters + string.digits + "-_"
+_KV = "= \t"
```
frostming: Can we only keep the `*_SET` variants？All `__contains__` check should work, too.

frostming: Or it's best to rename them without the _SET suffix.

tfoutrein:
Done — dropped the duplicate raw-string constants and kept only frozensets, renamed without the `_SET` suffix (your preferred option). The two `consume(...)` calls that need a plain string now pass a literal. No behaviour change; the full test suite + conformance still pass.

Note on CI: all unit-test jobs (every OS × Python) + pre-commit + the `poetry` integration are green. The two `poetry-core` integration jobs failed at `poetry install` with a vendoring/lockfile resolution error ("This is likely not a Poetry issue…") — unrelated to this PR (it passed on `master` and on the Windows `poetry-core` job). I don't have permission to re-run them on this repo, but a re-run should clear it.

frostming (approved):

PR #498, opened 2026-06-09
Fix ParseError when a sub-table extends an array of tables out of order

### LINKED ISSUE #261: Parser sometimes rejects subtable of table array
The tomlkit parser rejects the following document:
```
[[fruit]]  # the issue disappears if [fruit] is a table instead of a table array
apple.color = "red"

[potato]  # the issue disappears if this line is removed

[fruit.apple.texture]
smooth = true
```
and raises:
```
tomlkit.exceptions.ParseError: Key "fruit" already exists. at line 7 col 0
```

It seems clear to me that the document is valid TOML 1.0.0. In any case the presence of the `[potato]` table should not make a difference wrt its validity.

This issue applies to the latest commit 6512eaa3937a6eb2136face6a6704a52aa33e6e3 on master as well as the latest release v0.11.6.

I'm experimenting with a TOML fuzzer which revealed this issue. I guess most normal applications are not likely to hit this case, so I understand if this gets low priority. It still seems worthwhile to fix it though.

sarathfrancis90:
Fixes #261.

`tomlkit` raises `ParseError: Key "fruit" already exists.` on valid TOML when a sub-table header extends an array of tables with an unrelated table in between:

```toml
[[fruit]]
apple.color = "red"

[potato]

[fruit.apple.texture]
smooth = true
```

The contiguous form (without `[potato]`) already parses, because `[fruit.apple.texture]` gets consumed while the array element is still being parsed. Once an unrelated table breaks that up, the header is rebuilt as a fresh super table named `fruit` and appended to the document, colliding with the existing array.

A sub-table header naming an existing array of tables refers to the array's last element, so I merge its contents there instead of treating it as a redefinition. Defining a real table with an array's name (`[fruit]` after `[[fruit]]`) is still a genuine conflict and keeps raising.

One caveat worth flagging: on the round trip the out-of-order sub-table is normalized into the array element rather than kept at its original position. Byte-for-byte preservation isn't reachable here the way it is for plain tables — those lean on `OutOfOrderTableProxy`, which is dict-shaped, whereas an array element has to round-trip back as a list. The data is correct and stable on re-dump; happy to dig into preserving the exact layout if you'd prefer.

Added a regression test; full suite, ruff and mypy all pass.


dimbleby:
> One caveat worth flagging: on the round trip the out-of-order sub-table is normalized into the array element rather than kept at its original position

Seems like this would be worth a note in the readme or something. It is a break from the promise currently made:

> includes a parser that preserves all comments, indentations, whitespace and internal element ordering

Though, tomlkit would not be the only one - cf https://github.com/toml-rs/toml/tree/main/crates/toml_edit#limitations

sarathfrancis90:
Good point — added a Limitations note to the README spelling this out. As you say, it's the same trade-off `toml_edit` documents: the data is preserved, only the physical placement of that one out-of-order sub-table changes on re-serialization.

PR #499, opened 2026-06-09
Fix doubled comma when inserting into a comma-first array

### LINKED ISSUE #494: tomlkit inserts an invalid comma when appending to unusually formatted array
Bit of an edge case: but I found it so I might as well tell you about it.
```python
import tomlkit

src = """
 a = [
       1 # comma is on the next line
      ,2
     ]
"""

doc = tomlkit.parse(src)
assert doc == {"a": [1, 2]}, "src is valid TOML"

doc["a"].append(99)

out = tomlkit.dumps(doc)
print(out)
```
output:
```
 a = [
       1 # comma is on the next line
      ,2,
      ,99
     ]
```
which has too many commas between the 2 and the 99.

gaoflow:
Fixes #494.

In a comma-first formatted array the comma that separates an item from the previous one lives in that item's *indent* whitespace (the parser folds `\n ,` into the next group's indent when a comment ends the previous group). `Array.insert()` copies such an indent onto the new item, but then also adds a trailing comma to the previous item — and for middle inserts gives the new item a trailing comma too — so the output contains double separators and no longer parses:

```python
doc = tomlkit.parse("a = [\n      1 # one\n     ,2\n    ]\n")
doc["a"].append(99)
tomlkit.parse(tomlkit.dumps(doc))  # UnexpectedCharError: ',' — dumped "2,\n     ,99"
```

The fix skips both extra commas when the indent copied onto the new item already carries one: the embedded comma is the separator. Output is style-preserving (`,2` → `,99`) and round-trips. Same structure as the `__delitem__` `comma_in_indent` handling from #486, which `insert()` never got.

Regression tests cover append, middle insert (both previously produced unparseable output), and front insert (comma-less indent, unchanged behavior).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fix invalid serialization with a duplicated comma when removing a non-edge element from a parsed inline table. ([#486](https://github.com/python-poetry/tomlkit/pull/486))
+- Fix invalid serialization with a duplicated comma when appending or inserting into a comma-first formatted array. ([#499](https://github.com/python-poetry/tomlkit/pull/499))
 
 ## [0.15.0] - 2026-05-10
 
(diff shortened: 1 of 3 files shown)


frostming:
Thanks

PR #503, opened 2026-06-11
Merge super tables in place instead of deep-copying on every merge

tfoutrein:
## What

When a super table is merged into an existing one, `Container.append` deep-copied the existing target before appending the incoming branch:

```python
current = copy.deepcopy(current)
for k, v in item.value.body:
    current.append(k, v)
self._body[idx] = (key, current)
```

That copy is `O(size of current)`, and because the target accumulates every sibling that is merged into it, parsing a document with many subtables under a shared super table becomes **O(n²)** in the number of siblings. This is common in generated config / lock-style files with headers like `[a.b.c]`, `[a.b.d]`, `[a.b.e]`, …

Merging in place is `O(1)` per merge, so the parse is `O(n)` overall.

The deep copy only existed to stop the **out-of-order validation** pass from mutating the live tables it walks (`OutOfOrderTableProxy.validate` builds a temp container by appending those tables' bodies, which itself merges them). The isolating copy is therefore moved into that consumer, where it runs once for genuinely out-of-order tables instead of on every in-order merge.

## Benchmark

Parsing `n` subtables under a shared super table (`[s.r0.…]` … `[s.rN.…]`), median:

| n | before | after | speedup |
|---:|---:|---:|---:|
| 20 | 19 ms | 2.0 ms | ×9.7 |
| 80 | 300 ms | 8.2 ms | ×36 |
| 160 | 1243 ms | 16.5 ms | ×75 |
| 320 | 5326 ms | 34 ms | ×157 |

The growing speedup is the O(n²)→O(n) change. A realistic mixed document (`data.toml`) parses ~12% faster; nothing regresses.

## Tests

Full suite incl. the toml-test conformance submodule passes. No public API or behaviour change — round-trip output is byte-identical to `master`, verified by a differential over many nested / out-of-order / dotted-key / AoT documents (including `.add()` mutations on nested tables).


frostming (approved):

frostming:
Good job. Merged

PR #507, opened 2026-06-12
Merge AoT fragments when building an out-of-order table proxy

### LINKED ISSUE #505: KeyAlreadyPresent when accessing valid TOML with repeated array tables under an out-of-order table
### Description

`tomlkit.parse()` accepts a TOML document, but later accessing a parsed table with `doc.get("hooks")` can raise:

```text
tomlkit.exceptions.KeyAlreadyPresent: Key "Stop" already exists.
```
This happened with a valid TOML config that contains repeated array tables such as [[hooks.Stop]] plus sibling/subtables under [hooks].

The failure appears to happen while constructing an OutOfOrderTableProxy, not during parsing.

Observed behavior
This code path fails:
```python
import tomlkit
from pathlib import Path

text = Path("config.toml").read_text(encoding="utf-8")

doc = tomlkit.parse(text)      # succeeds
hooks = doc.get("hooks")       # raises KeyAlreadyPresent
```
Traceback:

```python
  File "<string>", line 11, in <module>
  File "<frozen _collections_abc>", line 807, in get
  File ".../tomlkit/container.py", line 711, in __getitem__
    item = self.item(key)
  File ".../tomlkit/container.py", line 517, in item
    return OutOfOrderTableProxy(self, idx)
  File ".../tomlkit/container.py", line 912, in __init__
    self._internal_container._raw_append(k, v)
  File ".../tomlkit/container.py", line 360, in _raw_append
    raise KeyAlreadyPresent(key)
tomlkit.exceptions.KeyAlreadyPresent: Key "Stop" already exists.
```
### Expected behavior
If tomlkit.parse() accepts the TOML document, then accessing doc.get("hooks") should not fail with KeyAlreadyPresent.

Repeated array tables like this are valid TOML:
```toml
[hooks]

[[hooks.Stop]]
matcher = ".*"

[[hooks.Stop.hooks]]
type = "command"
command = "one"
# even some actual content in here is valid but tomlkit gives an issue if something else is present
[[hooks.Stop]]
matcher = ".*"

[[hooks.Stop.hooks]]
type = "command"
command = "two"

[hooks.state]

[hooks.state.example]
trusted_hash = "sha256:abc"
```
Actual behavior
On the original config, tomlkit.parse(text) succeeded, but doc.get("hooks") raised KeyAlreadyPresent: Key "Stop" already exists.

Versions tested
This reproduces with:
```text
Python 3.12.7
tomlkit 0.14.0
tomlkit 0.15.0
tomlkit 0.15.0 produced the same class of traceback:

0.15.0
tomlkit.exceptions.KeyAlreadyPresent: Key "Stop" already exists.
```
Notes
The issue seems related to `OutOfOrderTableProxy` handling repeated array tables under a table that also contains sibling tables/subtables. Reordering the TOML so repeated [[hooks.Stop]] entries are grouped before other sibling tables avoids the failure in my application, but the original TOML should still be accessible after parsing.

gaoflow:
Fixes #505.

When an array of tables is split across out-of-order parts of the same table, `OutOfOrderTableProxy.__init__` fed both fragments to `_raw_append`, which only knows how to chain duplicate `Table` parts and raised `KeyAlreadyPresent`. Since the proxy is also built by `__contains__`, valid documents (accepted by `tomllib`) failed at parse time once a `[hooks.*]` header followed the second fragment.

The fragments are now presented as a new merged `AoT` that references the live element tables — neither fragment is mutated, so the table parts keep rendering their own elements (round-trip stays byte-identical) and element-level mutations through the proxy still write through to the document. This mirrors the isolation note in `OutOfOrderTableProxy.validate`, which deep-copies for the same reason.

The reported config only reproduces when the repeated AoT is genuinely out of order (split by an unrelated table) — contiguous repeats as literally pasted in the issue already worked; verified the merged result matches `tomllib` for both shapes.

### DIFF
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -1004,7 +1004,11 @@ class OutOfOrderTableProxy(_CustomDict):  # type: ignore[type-arg]
                 self._tables.append(_item)
                 table_idx = len(self._tables) - 1
                 for k, v in _item.value.body:
-                    self._internal_container._raw_append(k, v)
+                    merged = self._merge_aot_fragment(k, v)
+                    if merged is None:
+                        self._internal_container._raw_append(k, v)
+                    else:
+                        v = merged
                     key_indices = self._tables_map.setdefault(k, [])  # type: ignore[arg-type]
                     if table_idx not in key_indices:
                         key_indices.append(table_idx)
@@ -1012,6 +1016,34 @@ class OutOfOrderTableProxy(_CustomDict):  # type: ignore[type-arg]
                         dict.__setitem__(self, k.key, v)
 
         self._internal_container._validate_out_of_order_table()
+
+    def _merge_aot_fragment(self, key: Key | None, item: Item) -> AoT | None:
+        """
+        Merge an array-of-tables fragment from a later out-of-order table part.
+
+        An AoT whose elements are split across out-of-order parts of the same
+        table arrives here once per part; ``_raw_append`` only knows how to
+        chain duplicate ``Table`` parts and would raise ``KeyAlreadyPresent``.
+        The fragments are presented as a new merged ``AoT`` referencing the
+        live element tables, without mutating either fragment (the parts keep
+        rendering their own elements).
+
+        Returns the merged ``AoT``, or ``None`` if this is not such a fragment.
+        """
+        internal = self._internal_container
+        if key is None or not isinstance(item, AoT) or key not in internal._map:
+            return None
+        idx = internal._map[key]
+        if isinstance(idx, tuple):
+            return None
+        existing = internal._body[idx][1]
+        if not isinstance(existing, AoT):
+            return None
+
+        merged = AoT([*existing.body, *item.body], parsed=True)
+        internal._body[idx] = (internal._body[idx][0], merged)
+        dict.__setitem__(internal, key.key, merged.value)
+        return merged
 
     def unwrap(self) -> dict[str, Any]:
         return self._internal_container.unwrap()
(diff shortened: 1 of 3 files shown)


tomlkit/container.py
```diff
@@ -1013,6 +1017,34 @@ def __init__(self, container: Container, indices: tuple[int, ...]) -> None:
...
+        live element tables, without mutating either fragment (the parts keep
+        rendering their own elements).
+
+        Returns the merged ``AoT``, or ``None`` if this is not such a fragment.
+        """
+        internal = self._internal_container
+        if key is None or not isinstance(item, AoT) or key not in internal._map:
+            return None
+        idx = internal._map[key]
+        if isinstance(idx, tuple):
```
frostming, line 1037: why skip? it could be separated for more than twice. As in L1044

binggao1230:
Good question. The ">2 fragments" case is actually handled by the single-index path right below, not the tuple branch.

When an array of tables is split across three or more out-of-order parts, each later fragment merges into the AoT that's already in `_body` at a single index: `_merge_aot_fragment` rebuilds the merged AoT in place and leaves `internal._map[key]` pointing at that same int index, so the next fragment takes the same path and appends to the growing element list. `idx` never becomes a tuple for genuine AoT fragments.

A tuple index only shows up when the key already resolved to several distinct body positions, which here means a degenerate collision (e.g. a non-AoT part sharing the key) where an in-place merge isn't well defined; returning `None` there preserves the existing `_raw_append` behavior instead of indexing `_body` with a tuple.

I pushed a clarifying comment on that branch and added `test_out_of_order_table_merges_three_aot_fragments`, which splits `[[hooks.Stop]]` across three out-of-order parts and asserts all three elements (`a`, `b`, `c`) round-trip in order.

frostming (approved):

frostming:
I now understand. I might have misjudged it before.

PR #510, opened 2026-06-12
Limit parser nesting depth instead of crashing with RecursionError

gaoflow:
Fixes #459 (picking this up since #468 went stale).

Crafted input could crash the process with a `RecursionError` via three unbounded recursion paths: nested arrays, nested inline tables, and dotted keys (every fragment recursed through `_parse_key`/`_parse_bare_key`, which the issue didn't mention but has the same effect — `'a.'*10000 + 'a = 1'` also crashed). Values nested more than 100 levels deep and keys with more than 100 dotted fragments now raise `ParseError` with position info; `MAX_NESTING_DEPTH` is a class attribute so subclasses can override it.

Dotted-key parsing is now a loop instead of recursion (the chain is linear, and `concat` flattening is associative so left-to-right building produces identical keys); the cap is still needed there because each fragment creates a level of nested containers and operations like `Container.parsing()` recurse over that tree. Documents at exactly the limit still parse and round-trip — covered in the tests, which all fail with `RecursionError` on master.


frostming (approved):

frostming:
Can you also add a bottom line mentioning which agent and model you are using, if any. This will help us evaluate the PR. Thanks!

PR #511, opened 2026-06-12
Prefix every line of a multiline comment with #

### LINKED ISSUE #449: Multiline comments
Currently this will not work:
```
        doc = document()
        doc.add(comment('A comment is here'))
        doc.add(comment("""
---------------------------------------------------------------------------------------------------------------------
A multi line
comment block is here
---------------------------------------------------------------------------------------------------------------------
"""))
```
It creates:

```
# A comment is here
#
---------------------------------------------------------------------------------------------------------------------
A multi line
comment block is here
---------------------------------------------------------------------------------------------------------------------

```

Would like it to make:
```
# A comment is here
# ---------------------------------------------------------------------------------------------------------------------
# A multi line
# comment block is here
# ---------------------------------------------------------------------------------------------------------------------
```
In api.py if def comment() is changed to:
```
def comment(string: str) -> Comment:
    """Create a comment item."""

    lines = string.splitlines()
    if len(lines) > 1:
        return Comment(Trivia(comment_ws="  ", comment='\n# '.join(lines)[1:]))
    else:
        return Comment(Trivia(comment_ws="  ", comment="# " + string))
```

It will the create comments as shown above.

Possibly include that in master?

Thanks,
Curits

gaoflow:
Fixes #449.

`comment("line one\nline two")` only prepended `# ` to the whole string, so every line after the first was emitted without a `#` and the document no longer re-parsed:

```
# line one
line two        # <- invalid
```

`comment()` now prefixes each line individually (a bare `#` for empty lines), so a multiline comment renders as a valid block of comment lines and round-trips. Single-line comments are unchanged.


frostming:
We hope that comments will always be single-line, and multi-line comments should be rejected, otherwise it will cause potential problems, such as:

```toml
a = 1
```
```python
doc["a"].comment("line1\nline2")
```
will produce:

```toml
a = 1 # line1
#line2
```
The second line should not belong to the table item any more.

So, the right approach is:
```python
doc["a"].comment("line1")
doc.append(comment("line2"))
```
Losing some flexibility, the code can more clearly reflect the actual structure.

frostming (approved):

PR #516, opened 2026-06-14
reject a table redefined after its parent table header

netliomax25-code:
## Summary

1. `Container.append` sends a concrete `[table]` header that lands inside an existing implicit/super table through `_validate_table_candidate`, which only flagged a child key on a type change or a dotted-key redefinition.
2. A child table header that was already defined fell through to a plain merge, so a subtable redefined after its parent received its own header (`[a.b]` / `[a]` / `[a.b]`, and the deeper `[a.b.c]` / `[a]` / `[a.b.c]`) was merged silently instead of rejected, while `tomllib` and the toml-test suite treat these as declaring the same table twice.
3. `_validate_table_candidate` now raises when the existing child and the candidate child are both concrete (non-super) tables, and recurses through matching super tables so a duplicate nested deeper is caught as well. Valid out-of-order tables (a new sibling subtable, or extending a table with a deeper header) keep parsing, and the existing tests plus the toml-test 1.0/1.1 suites pass.

## Agent Drafting Metadata

- Agent:
- Model:
- Notes:

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -112,3 +112,18 @@ def test_parser_rejects_tab_in_bare_key(content: str) -> None:
     parser = Parser(content)
     with pytest.raises(ParseError):
         parser.parse()
+
+
+@pytest.mark.parametrize(
+    "content",
+    [
+        "[a.b]\n[a]\n[a.b]",
+        "[a.b]\nx = 1\n[a]\n[a.b]\ny = 2",
+        "[a.b.c]\n[a]\n[a.b.c]",
+        "[a.b]\n[a.c]\n[a]\n[a.b]",
+    ],
+)
+def test_parser_rejects_table_redefined_after_parent(content: str) -> None:
+    parser = Parser(content)
+    with pytest.raises(ParseError):
+        parser.parse()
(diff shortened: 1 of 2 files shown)


frostming:
> 2\. while `tomllib` and the toml-test suite treat these as declaring the same table twice.

Hmm, we are using toml-test suite but why it isn't caught?

netliomax25-code:
The suite just does not have a case with this shape. Its table-redefinition cases all re-declare the parent/super header, not the child:

- `invalid/table/super-twice` is `[a.b]` / `[a]` / `[a]`
- `invalid/table/duplicate-key-09` is `[a]` / `[a.b]` / `[a]`
- `invalid/table/duplicate-key-01` is `[a]` / `[a]`

tomlkit already rejects all of those. The gap is when the child header is the repeated one with the parent declared in between, like `[a.b]` / `[a]` / `[a.b]`. That falls through to a plain merge: on master it parses to `{"a": {"b": {}}}`, and `[a.b]` / `x=1` / `[a]` / `[a.b]` / `y=2` silently drops `x=1`. tomllib rejects both.

Since toml-test is a submodule we do not own, I added the regression cases in `tests/test_parser.py`. Happy to send the input upstream to BurntSushi/toml-test too if you want it covered there.

frostming (approved):

PR #520, opened 2026-06-15
Return a single key from key() for a single-element list (fixes #430)

### LINKED ISSUE #430: Different behavior for SingleKey and DottedKey with single value
I am building a tomlkit document and want to generate keys dynamically. Sometimes the lengths of my key specifications are only 1 long, but tomlkit still generates a DottedKey out of it. The documentation of `tomlkit.key` is very clear about it: specifying a list will return a DottedKey.

But I think it would be a nice addition if the function checks the length of the list before making that decision.

A minimal example of what goes wrong now:

```py
import tomlkit

doc = tomlkit.document()
doc.append(tomlkit.key("my_key"), "value")
key_in_doc = "my_key" in doc # returns True

doc2 = tomlkit.document()
doc.append(tomlkit.key(["my_key"]), "value")
key_in_doc2 = "my_key" in doc # returns False

assert key_in_doc == key_in_doc2 # unfortunately not
```

The `in` functionality doesn't seem to recognize the DottedKey as equal to a single `str`.

I was able to do a workaround in the code now by:

```py
doc = tomlkit.document()
if len(key_spec) == 1:
    doc.append(tomlkit.key(key_spec[0]), "value")
else:
    doc.append(tomlkit.key(key_spec, "value")
```

Or in whatever way you want to implement it. At least, I really needed to unwrap the key_spec in order to make the `in` function work correctly.

Tested with tomlkit 0.13.3

tienvantranspk-tech:
Fixes #430.

## Problem
`key(["my_key"])` builds a single-element `DottedKey`, which behaves differently from `key("my_key")` (a `SingleKey`). For example:

    import tomlkit
    doc = tomlkit.document()
    doc.append(tomlkit.key(["my_key"]), "value")
    "my_key" in doc        # -> False  (key("my_key") gives True)

## Fix
In `key()`, when the iterable yields exactly one key, return that `SingleKey` instead of wrapping it in a `DottedKey`. Multi-element lists are unchanged (still dotted keys).

## Tests
Added a regression test. The existing suite passes (333 passed, excluding the `toml-test` git submodule which is not initialised in my checkout), and `ruff check` / `ruff format` are clean.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -365,6 +365,22 @@ def test_add_dotted_key() -> None:
     assert table.as_string() == "foo.bar = 1\n"
 
 
+def test_key_with_single_element_list_is_a_single_key() -> None:
+    # Regression test for
+    # https://github.com/python-poetry/tomlkit/issues/430: a one-element key
+    # list should behave like a plain key, not a single-element dotted key.
+    doc = tomlkit.document()
+    doc.append(tomlkit.key(["foo"]), "value")
+    assert "foo" in doc
+    assert doc["foo"] == "value"
+    assert doc.as_string() == 'foo = "value"\n'
+
+    # Multi-element key lists are still dotted keys.
+    doc = tomlkit.document()
+    doc.add(tomlkit.key(["foo", "bar"]), 1)
+    assert doc.as_string() == "foo.bar = 1\n"
+
+
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


frostming (approved):

tienvantranspk-tech:
Thanks for the quick review and merge, @frostming! Glad to help out with tomlkit.

Issue #513, opened 2026-06-12
an emptied dotted-key captures the elements that follow it


dimbleby:
```python
import tomlkit

text = """\
a.b = 1
c.d = 2
"""
doc = tomlkit.loads(text)

doc["a"] = {}

output = tomlkit.dumps(doc)
print(output)
```
result
```toml
[a]

c.d = 2
```
in which `c.d` has wrongly become a child of `a`

PR #528, opened 2026-06-18
Fix stale dotted-key prefix when replacing a table value with a plain dict

### LINKED ISSUE #524: Assigning a dict over a dotted key renders a stale parent. prefix
Similar flavour to #513, but this one introduces an unwanted header where that one only failed to reorder the document.  Maybe they will turn out to have the same fix.
```python
import tomlkit

doc = tomlkit.loads("fruit.apple = true\n")
doc["fruit"] = {"a": 1}

print(tomlkit.dumps(doc))
```
result
```
[fruit]
fruit.a = 1
```
which has too many fruit.

gaoflow:
Fixes #524.

After parsing `fruit.apple = true`, the document-body key `fruit` retained `_dotted=True` from `_handle_dotted_key`. When the whole table is later replaced via `doc['fruit'] = {'a': 1}`, `_replace_at` reuses the old key object (same name + same kind), keeping the stale flag. This causes rendering to both emit a `[fruit]` header (new table is not a super-table) and prefix every child with `fruit.` (because `key.is_dotted()` is true), producing:

```toml
[fruit]
fruit.a = 1
```

instead of the correct `a = 1`.

The fix clears `_dotted` on the reused key in `_replace_at`. The dotted heritage belongs to the original parse — a full table replacement gets fresh rendering behaviour. Incremental modifications (like `doc['fruit']['banana'] = 2`) are unaffected because they only reach the table-level `_replace_at`, not the document-level one that reuses the dotted-marked outer key.


binggao1230:
Superseded by #526 which fixes both #524 and #513 with a more comprehensive approach.

PR #529, opened 2026-06-18
Fix KeyAlreadyPresent error when AoT is extended by a dotted-key header after an unrelated table

### LINKED ISSUE #261: Parser sometimes rejects subtable of table array

gaoflow:
Fixes #261.

When a dotted-key header like `[fruit.apple.texture]` appears after an unrelated table following an `[[fruit]]` array-of-tables, the parser creates a super-table chain that `Container.append` receives with the same root key (`fruit`) as the existing AoT. Since the new item is a table (not an explicit AoT element), append previously raised `KeyAlreadyPresent`.

Example that was rejected before this fix:

```toml
[[fruit]]
apple.color = "red"

[potato]

[fruit.apple.texture]
smooth = true
```

The fix walks the super-table produced by the parser and does a deep merge of its children into the last element of the existing AoT — the TOML spec implicitly places such dotted-key extensions under the most recent table/array-table header.

The rendering order of unrelated tables interspersed with AoT extensions is not preserved (this is a pre-existing style-preservation limitation), but the parsed data is semantically correct and round-trips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Fix `comment()` producing invalid TOML for a multiline string by prefixing every line with `#`, not just the first. ([#449](https://github.com/python-poetry/tomlkit/issues/449))
 - Fix the separator comma being swallowed by a trailing comment when appending a key to a multiline inline table, leaving the new key without a separator so the result no longer round-trips. ([#512](https://github.com/python-poetry/tomlkit/issues/512))
 - Fix a `KeyAlreadyPresent` error when parsing or accessing an out-of-order table whose array-of-tables elements are split across the table's parts. ([#505](https://github.com/python-poetry/tomlkit/issues/505))
+- Fix a `KeyAlreadyPresent` error when a dotted-key header extension (e.g. `[fruit.apple.texture]`) appears after an unrelated table following an array-of-tables (e.g. `[[fruit]]`). ([#261](https://github.com/python-poetry/tomlkit/issues/261))
 
 ## [0.15.0] - 2026-05-10
 
--- a/README.md
+++ b/README.md
@@ -26,6 +26,15 @@ Part of the implementation has been adapted, improved and fixed from [Molten](ht
 
 See the [documentation](https://tomlkit.readthedocs.io/) for more information.
 
+## Limitations
+
+`tomlkit` preserves the layout of parsed documents where it can. One known
+exception is an out-of-order sub-table that extends the last element of an array
+of tables, such as `[fruit.apple.texture]` appearing after an unrelated table
+following `[[fruit]]`. When the document is serialized again, that sub-table is
+emitted inside the array element it belongs to. Its data is preserved, but its
+physical position in the file is normalized.
+
 ## Installation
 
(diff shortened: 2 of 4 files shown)


dimbleby:
Is this better or worse than #498?  Why?

binggao1230:
Thanks for pointing at #498. The semantic direction is the same, but this branch is intended to be a stricter superset: #498 appends the immediate super-table children into the last AoT element, while this branch recursively merges super-table children, so it also keeps existing nested table contents such as `apple.texture.rough = false` followed later by `[fruit.apple.texture]`.

While checking that comparison, I found this branch was too permissive for invalid redefinitions: `apple.texture = "rough"` followed by `[fruit.apple.texture]` could replace the scalar with a table. I pushed 2b5f890 to keep those as parse errors and added tests for both the recursive merge case and the redefinition-error cases.

So compared with #498: better on recursive merge coverage after the latest commit, similar caveat on layout normalization. #498 also adds a README limitations note for that caveat; I can add the same note here if you would prefer this PR to carry the documentation change too.

Validation run locally: `pytest -q` (1027 passed), plus `ruff check` and `ruff format --check` on the changed files.

dimbleby:
> I can add the same note here if you would prefer this PR to carry the documentation change too

I am just a passer by, a maintainer will have to decide which (if either) pull request to accept and what the docs should say.

But IMO if code is knowingly breaking tomlkit's promises about preserving element ordering - that wants writing down.

binggao1230:
I agree. I added a short README limitation note in 854c2fb documenting that this out-of-order AoT sub-table case preserves the data but normalizes the sub-table's physical position when serialized.

Checked locally:
- `pytest -q tests/test_toml_document.py`
- `ruff check tomlkit/container.py tests/test_toml_document.py`
- `ruff format --check tomlkit/container.py tests/test_toml_document.py`
- `git diff --check`

frostming:
#507 and #498 are also fixing this, @gaoflow please review if anything is not covered there.

binggao1230:
Thanks for pointing me to those. I reviewed #507 and #498: #498 covers the same out-of-order AoT sub-table case from #261, including the `[fruit.apple.texture]` after an unrelated table scenario, and #507 covers the separate split AoT fragment case from #505. Since both are merged now, this PR is duplicate/stale. Closing it.

PR #537, opened 2026-06-22
Preserve table display names on self assignment

### LINKED ISSUE #291: Difference in whitespace in dot notation disappears
I'm using whitespace to vertically align some toml config and when editing said config with your great library noticed that alignment disappears, below is a small repro

```py
content = """
[keys.a]
#    ↓ this whitespace disappears
[keys .'a'.'c']
  'd'	= 'e'
"""
def testtoml():
  from tomlkit import dumps
  from tomlkit import parse
  doc = parse(content)
  for mode in doc["keys"]: # self to self shouldn't change anything?
    doc["keys"][mode] = doc["keys"][mode]
  assert dumps(doc) == doc
  # print(dumps(doc))
testtoml()
```

It seems that the latter whitespace matches the first encountered in `[keys.a]` (if I add whitespace to `[keys.a]` then it propagates to further as well

Is it possible to preserve the difference in whitespace so that editing wouldn't "import" whitespace from upper keys, but retain its own individual?

(tested this on a master branch of your library)

gaoflow:
Fixes https://github.com/python-poetry/tomlkit/issues/291.

## Summary
- avoid invalidating parsed table display names when a table is assigned back to its existing slot
- add a regression test for preserving whitespace and quoting in dotted table headers after self-assignment

## Verification
- uv run --no-project --with pytest --with pyyaml python -m pytest tests/test_toml_document.py::test_replace_table_with_itself_preserves_display_name tests/test_toml_document.py::test_replace_with_table_of_nested tests/test_toml_document.py::test_nested_table_update_display_name -q
- uv run --no-project --with pytest --with pyyaml python -m pytest -q
- uv run --no-project --with ruff ruff check tomlkit/container.py tests/test_toml_document.py
- uv run --no-project --with ruff ruff format --check tomlkit/container.py tests/test_toml_document.py
- python -m compileall tomlkit/container.py tests/test_toml_document.py
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Fix a `KeyAlreadyPresent` error when parsing or accessing an out-of-order table whose array-of-tables elements are split across the table's parts. ([#505](https://github.com/python-poetry/tomlkit/issues/505))
 - Out-of-order value-vs-table and dotted-key-vs-table redefinitions are now rejected at parse time instead of being silently accepted or raising only on access. The parser also detects when a non-dotted key is a prefix of an existing dotted key, matching the stdlib `tomllib` behaviour. ([#523](https://github.com/python-poetry/tomlkit/issues/523))
 - Reject tables inserted into inline tables instead of serializing invalid TOML. ([#531](https://github.com/python-poetry/tomlkit/issues/531))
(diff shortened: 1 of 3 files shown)


frostming:
Please resolve conflict and add changelog

binggao1230:
Done — rebased onto master (the conflict was just two adjacent new test functions; kept both) and added a changelog entry under Fixed. The bug still reproduces on current master: `doc[k] = doc[k]` normalizes `[keys .'a'.'c']` to `[keys.a.'c']`, and this restores exact header preservation. Full suite green.

PR #538, opened 2026-06-22
reject overlong decimal integer instead of coercing to inf

netliomax25-code:
## Summary

1. `_parse_number` runs `int(sign + clean, base)` inside a `try/except ValueError` that, on failure, falls back to `float(sign + clean)`.
2. A decimal integer literal with more digits than Python's int-from-string limit (4300 by default) makes `int()` raise `ValueError`, so the literal is parsed as a float instead. A long run of digits ends up as `inf`, so `a = 9999...` (4301+ digits) silently returns a float `inf` rather than being rejected as an out-of-range integer.
3. Restrict the float fallback to literals that are actually float-shaped (a fractional dot, a base-10 exponent, or inf/nan). An integer that `int()` cannot parse now returns `None`, so the parser raises `InvalidNumberError`. Hex/oct/bin literals (no digit limit) and all valid float literals are unaffected.

Repro: `tomlkit.parse("a = " + "9" * 4301)["a"]` returns float `inf` before the change; it raises `InvalidNumberError` after. The 4300-digit value still parses as an `Integer`.

## Agent Drafting Metadata

- Agent:
- Model:
- Notes:

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from tomlkit.exceptions import EmptyTableNameError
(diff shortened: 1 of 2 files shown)


netliomax25-code:
Pushed a fix for the macOS/Windows 3.9 failures: those runners are on 3.9.13, which predates the int-from-string digit limit (backported in 3.9.14), so the new test is now skipped there. The parser change itself is a no-op on those versions since int() never raises.

frostming (approved):

PR #539, opened 2026-06-22
escape or reject a bare carriage return in multiline strings

netliomax25-code:
## Summary

1. `String.from_raw` builds a multiline basic string by escaping every control character except a line feed and a carriage return, so a lone CR (one that is not the CR of a CRLF pair) is written out raw.
2. The multiline literal path drops that same CR from `invalid_sequences`, so a lone CR there is neither escaped nor rejected.
3. Either way `tomlkit.string(value, multiline=True)` and its literal variant emit a document with a raw lone CR, which the tomlkit parser and the stdlib `tomllib` both reject as a control character, so the value stops round-tripping.
4. Escape a lone CR as `\r` for multiline basic strings, and raise `InvalidStringError` for multiline literal strings; LF and CRLF line endings are left untouched.

## Agent Drafting Metadata

- Agent: Claude Code
- Model: Claude (Anthropic)
- Notes: AI tooling helped find the bug and draft the fix and tests. The root cause, the change, and the tests were reviewed and validated by the author before submission (the new cases fail on the unpatched tree, and the full suite passes with the patch), and the author is responsible for what is submitted.

### DIFF
--- a/tomlkit/items.py
+++ b/tomlkit/items.py
@@ -234,7 +234,7 @@ def item(value: Any, _parent: Item | None = None, _sort_keys: bool = False) -> I
 
 # A carriage return that is not the CR of a CRLF line ending. Inside a multiline
 # string only a raw line feed or a CRLF pair is a valid line ending; a lone CR is
-# a control character. Literal strings cannot escape it, basic strings must.
+# a control character that a basic string must escape.
 _BARE_CR = re.compile(r"\r(?!\n)")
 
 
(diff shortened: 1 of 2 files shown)


netliomax25-code:
any update?

tomlkit/items.py
```diff
@@ -2239,9 +2245,15 @@ def from_raw(
         if any(c in value for c in invalid):
             raise InvalidStringError(value, invalid, type_.value)
 
+        if type_ is StringType.MLL and _BARE_CR.search(value):
+            raise InvalidStringError(value, {"\r"}, type_.value)
+
```
frostming: Is it possible to merge it to the if-statement above? It’s not ideal to check for a single `\r` separately.

netliomax25-code: Done. Kept `\r` in the MLL invalid set and collapsed CRLF pairs to `\n` before the check, so a lone CR now trips the existing if-statement instead of a separate one. Full suite still passes.

frostming:
Conflicts to be resolved

netliomax25-code:
Rebased onto master, conflicts resolved. The conflict was with the leading-newline fix from #551 in `from_raw`; kept both changes, with the bare-CR escape applied before the leading-newline check so an escaped CR doesn't get a spurious extra newline. Full suite passes (1055).

davidpavlovschi (approved):
The bare-CR check now distinguishes CRLF line endings from lone carriage-return control characters in both multiline forms. I verified the focused string API suite (42 tests) and 780 generated combinations of text, LF, CRLF, and bare CR: multiline basic strings round-trip through tomlkit and remain valid for `tomllib`, while multiline literals reject exactly the values containing a CR outside a CRLF pair. The leading-newline behavior from #551 remains intact.

PR #544, opened 2026-06-24
reject a comment containing line breaks in Item.comment and add_line

netliomax25-code:
## Summary

1. `Item.comment()` and `Array.add_line(comment=...)` accepted a comment containing an embedded newline and rendered everything after the first line without a `#`, so the output no longer re-parses (tomlkit and `tomllib` both reject it).
2. Per review these two APIs are single-line only, so a comment containing a line break now raises `ValueError` up front instead of being reformatted; the item is left unmodified when it raises.
3. Single-line comment behavior is unchanged. Tests cover both rejection paths; full suite passes.

## Agent Drafting Metadata

- Agent:
- Model:
- Notes:

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -515,6 +515,13 @@ def test_array_add_line() -> None:
     )
 
 
+def test_array_add_line_multiline_comment_is_rejected() -> None:
+    t = api.array()
+    with pytest.raises(ValueError, match="line breaks"):
+        t.add_line(1, 2, 3, comment="first line\nsecond line")
+    assert t.as_string() == "[]"
+
+
 def test_array_add_line_invalid_value() -> None:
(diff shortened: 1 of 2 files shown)


netliomax25-code:
any update?

frostming:
They are different thing, `add_line()` and `item.comment()` are expected to accept single line comment ONLY. In case of string containing line breaks, you should reject it.

netliomax25-code:
Makes sense, reworked it that way:

1. `Item.comment()` and `add_line(comment=...)` now raise `ValueError` when the comment contains a line break, instead of trying to reformat it.
2. Single-line behavior is unchanged; tests updated to check the rejection and the full suite still passes.

Updated the title and description to match.

PR #549, opened 2026-06-30
Fix array of tables replacing a dotted key swallowing the next sibling (#542)

### LINKED ISSUE #542: AoT captures too many keys
```python
#!/usr/bin/env python3

import tomlkit


doc = tomlkit.parse("a.b = 1\nc.d = 2\n")
aot = tomlkit.aot()
t = tomlkit.table()
t["x"] = 9
aot.append(t)
doc["a"] = aot
print(tomlkit.dumps(doc))
```
result
```
[[a]]
x = 9
c.d = 2
```
in which `c.d` has wrongly been captured by the array of tables.

Similar in flavour to other recent bug reports and fixes, I expect the fix will be nearby too.

HarperZ9:
Fixes #542.

Assigning an array of tables over a dotted key swallows the following sibling on round-trip:

```python
import tomlkit
doc = tomlkit.parse("a.b = 1\nc.d = 2\n")
aot = tomlkit.aot()
t = tomlkit.table(); t["x"] = 9
aot.append(t)
doc["a"] = aot
print(tomlkit.dumps(doc))
```

Before:
```
[[a]]
x = 9
c.d = 2
```
`c.d` is captured by `[[a]]`, so it round-trips to `{"a": [{"x": 9, "c": {"d": 2}}]}` and the top-level `c` is lost.

### Cause
In `Container._replace_at`, the `dotted_to_header` guard recognises only non-super `Table`s, not `AoT`s. That guard drives the "drop the dotted prefix and move the new header past the inline entries (plain values and dotted keys) it would otherwise capture" path that already fixes the table case (#513 / #524). Because an `AoT` never matched it, replacing a dotted key with an array of tables left the `[[a]]` header in the dotted key's inline slot, capturing the next dotted sibling.

### Fix
Extend `dotted_to_header` to also cover `AoT` (an array of tables always renders with its own `[[header]]`, so it needs the same treatment as a non-super table). One-line logic change.

After:
```
c.d = 2

[[a]]
x = 9
```
Round-trips to `{"c": {"d": 2}, "a": [{"x": 9}]}`.

Adds `test_replace_dotted_key_with_aot_keeps_following_sibling` (mirrors the existing #513 table test) and a changelog entry. Full suite passes; `ruff check` / `ruff format` are clean.


HarperZ9:
Fresh local verification from a Windows checkout of head `b39439a49b9c03d25ca5627d5bcd04ccfdfed77e` with Python 3.12.10:

- `uv run --with pytest --with pytest-cov python -m pytest tests/test_toml_document.py::test_replace_dotted_key_with_aot_keeps_following_sibling tests/test_toml_document.py::test_replace_dotted_key_with_table_keeps_following_sibling tests/test_toml_document.py::test_replace_value_with_table_keeps_following_dotted_sibling -q` -> 3 passed
- `uv run --with ruff ruff check tomlkit/container.py tests/test_toml_document.py` -> all checks passed
- `uv run --with ruff ruff format --check tomlkit/container.py tests/test_toml_document.py` -> 2 files already formatted

This keeps the local verification focused on the AoT replacement regression and the adjacent dotted-key replacement paths. The checkout is clean after the run.

PR #552, opened 2026-07-05
Preserve inline table key order in item() (#546)

### LINKED ISSUE #546: item() reorders keys (dict-valued before scalar) even with the default sort_keys=False

uttam12331:
## Summary

Closes #546.

`item()` reordered the keys of an inline table under the default `sort_keys=False`:

```python
tomlkit.item([[{"a": {"x": 1}, "b": 2}]]).as_string()
# -> '[[{a = {x = 1}, b = 2}]]'   (before: '[[{b = 2, a = {x = 1}}]]')
```

## Root cause

In the `list`/`tuple` branch of `item()`, the sort key was

```python
key=lambda i: (isinstance(i[1], dict), i[0] if _sort_keys else 1)
```

so `isinstance(i[1], dict)` stayed in the tuple **unconditionally**, forcing dict-valued keys last even when not sorting. The sibling `dict` branch has the correct form (`(isinstance(i[1], dict), i[0]) if _sort_keys else 1`).

## Fix

Keep the dicts-last ordering **only for arrays of tables** (`Table` constructor), which render as `[[table]]` headers where a dict-valued key would otherwise capture the following keys — this is why simply removing the dicts-last ordering (PR #547) was not acceptable. Inline tables cannot capture, so their insertion order is preserved unless `sort_keys=True` is passed (matching the `dict` branch).

Verified:
- inline: `[[{a = {x = 1}, b = 2}]]` (order preserved)
- inline with `_sort_keys=True`: `[[{b = 2, a = {x = 1}}]]` (still sorted, dicts-last)
- array of tables: unchanged — `b = 2` still emitted before the `[a]` sub-table header

## Tests

Added `test_item_inline_table_preserves_key_order` covering all three cases. It fails on `master` and passes with this change. Full suite passes (356 passed); `ruff check`/`ruff format` clean; added a CHANGELOG entry.


tomlkit/items.py
```diff
@@ -161,14 +161,22 @@ def item(value: Any, _parent: Item | None = None, _sort_keys: bool = False) -> I
...
+        if table_constructor is Table:
+            # AoT tables render as ``[[table]]`` headers, so dict-valued keys
+            # must come last to avoid them capturing the following keys.
+            def _sort_key(i: tuple[Any, Any]) -> Any:
+                return (isinstance(i[1], dict), i[0] if _sort_keys else 1)
+        else:
+            # Inline tables cannot capture, so preserve insertion order unless
+            # explicitly sorting (matching the dict branch above). See #546.
+            def _sort_key(i: tuple[Any, Any]) -> Any:
+                return (isinstance(i[1], dict), i[0]) if _sort_keys else 1
```
dimbleby: why are you keeping dictionaries last in this branch?

uttam12331: Only for consistency with the sibling `dict` branch above (line ~146), which uses this exact sort key — `(isinstance(i[1], dict), i[0]) if _sort_keys else 1` — and also constructs `InlineTable`s (when `_parent` is an `Array`/`InlineTable`). With this form, `sort_keys=True` produces the same ordering for an inline table regardless of whether it was built via the dict branch or the list branch.

I kept the scope of this PR to the `sort_keys=False` default (the #546 bug). If you'd rather have sorted inline tables be purely alphabetical (no dicts-last), I'm happy to change it — but then the dict branch should probably change too, so the two paths don't diverge. Let me know which you'd prefer.

dimbleby: The sibling branch needs to put dictionaries last so as to avoid unwanted scalar capture.  

The whole point of your fix is that the inline-table branch has no such requirement, so it seems weird that you would interpret sort_keys to mean "sort keys, but still put dictionaries last"

uttam12331: Fair point — dicts-last is capture-avoidance, not a sorting preference, and inline tables don't need it. Changed in 4f3210e: the inline-table branch now sorts purely by key when `sort_keys=True` (and still preserves insertion order by default). Test updated accordingly; full suite passes (356).

davidpavlovschi (approved):
The split sort policy matches the rendering semantics: arrays of tables keep dict-valued children after scalars to avoid header capture, while inline tables preserve insertion order by default and sort only by key when requested. I exercised all three paths and ran the complete suite on this head: 1,036 passed.

PR #558, opened 2026-07-09
Fix promoted dotted-key header capturing following siblings (#556)

### LINKED ISSUE #556: updating a table captures too many keys
```python
#!/usr/bin/env python3

import tomlkit

doc = tomlkit.loads("a.b = 1\na.c = 2\na.d = 3\n")
doc["a"]["b"] = {"x": 9}

output = tomlkit.dumps(doc)

print(output)
```
result
```
[a.b]
x = 9
a.c = 2
a.d = 3
```
in which `a.c` and `a.d` have become children of the `[a.b]` table

similar to #542, #543, but not fixed by PRs currently open for those so should be distinct.

apoorvdarshan:
## Summary

Fixes #556.

Replacing a dotted-key value with a table promotes the key to a `[table]`
header, and any dotted keys that render *after* that header get visually
captured as its children, so the document no longer round-trips.

Repro from the issue:

```python
import tomlkit

doc = tomlkit.loads("a.b = 1\na.c = 2\na.d = 3\n")
doc["a"]["b"] = {"x": 9}
print(tomlkit.dumps(doc))
```

Before this change:

```toml
[a.b]
x = 9
a.c = 2
a.d = 3
```

`a.c` / `a.d` sit under `[a.b]` and re-parse as `a.b.a.c` / `a.b.a.d`.
`a.b`/`a.c`/`a.d` are parsed as three separate root fragments of the dotted
key `a`; promoting `a.b` moves the `[a.b]` header ahead of the fragments that
still render as plain dotted keys. The same capture also happens for a
differently-named trailing dotted key (e.g. a following `p.q`). This is a
distinct defect from the open PRs for #542/#543.

After:

```toml
a.c = 2
a.d = 3

[a.b]
x = 9
```

## Fix

`Container.as_string()` now runs the top-level body through a new
`_render_ordering()` helper before emitting it. Within the leading run of
root entries (everything up to the first real `[table]` / array-of-tables
header), it stably moves every *promoted* dotted header after the inline
entries (plain values and scalar-rendering dotted keys) that follow it,
preserving relative order otherwise. A small `_dotted_fragment_emits_header()`
predicate decides whether a dotted super-table fragment renders a header or
just dotted keys.

The reordering only engages when a promoted header actually precedes an inline
entry it would capture, so ordinary documents and genuine `[table]` sections
are left byte-for-byte unchanged.

## Tests / Verification

- Added three regression tests in `tests/test_toml_document.py`
  (`test_replace_dotted_key_keeps_following_same_name_siblings`,
  `test_replace_middle_dotted_key_keeps_following_same_name_siblings`,
  `test_replace_dotted_key_keeps_following_other_dotted_key`). All three fail
  on `master` and pass with the fix.
- Full suite green: `1038 passed`.
- `ruff check`, `ruff format --check`, and `mypy` are clean on the changed
  files.
- Added a `CHANGELOG.md` entry under `[unreleased] / Fixed`.

## Agent Drafting Metadata

- Agent: Claude Code
- Model: Claude Opus 4.8
- Notes: Bug reproduced, fix and tests written, and full suite/lint/type-check
  run locally before opening.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Fix a `KeyAlreadyPresent` error when parsing or accessing an out-of-order table whose array-of-tables elements are split across the table's parts. ([#505](https://github.com/python-poetry/tomlkit/issues/505))
 - Out-of-order value-vs-table and dotted-key-vs-table redefinitions are now rejected at parse time instead of being silently accepted or raising only on access. The parser also detects when a non-dotted key is a prefix of an existing dotted key, matching the stdlib `tomllib` behaviour. ([#523](https://github.com/python-poetry/tomlkit/issues/523))
 - Reject tables inserted into inline tables instead of serializing invalid TOML. ([#531](https://github.com/python-poetry/tomlkit/issues/531))
+- Fix replacing a dotted-key value with a table (e.g. `doc["a"]["b"] = {...}` where `a.b`/`a.c`/`a.d` were parsed as separate root fragments) capturing the sibling dotted keys that follow it. Promoting `a.b` to a `[a.b]` header used to leave the trailing root dotted keys (whether the same prefix `a.c`/`a.d` or a different one such as `p.q`) rendered after the header, so they were re-parsed as its children. Those inline entries now render before the promoted header, preserving the round-trip. ([#556](https://github.com/python-poetry/tomlkit/issues/556))
 
 ## [0.15.0] - 2026-05-10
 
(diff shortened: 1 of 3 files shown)


frostming:
Please check existing PR #550 before submitting new

apoorvdarshan:
Thanks for pointing this out. I reviewed #550 and agree that it overlaps substantially in the same container-ordering area. Although #556 exercises a sibling-replacement variant, keeping a competing implementation open would duplicate the ongoing work, so I’m closing this PR in favor of #550.

Issue #553, opened 2026-07-05
tomlkit does not serialize some empty arrays and dictionaries


dimbleby:
```python
#!/usr/bin/env python3

import tomlkit


doc = tomlkit.loads("""
x.y.z = 1

[[a]]
x = 1
""")

doc["a"].pop()
doc["x"]["y"].pop("z")

print(f"{doc=}")

rendered = tomlkit.dumps(doc)
print(f"{rendered=}")
```
result
```
doc={'x': {'y': {}}, 'a': []}
rendered='\n\n'
```
Two newlines is not a good TOML representation of that dictionary

PR #567, opened 2026-07-19
Render an emptied array of tables as an empty array

dchaudhari7177:
## Summary

Addresses the array-of-tables half of #553.

When every element is removed from an array of tables, there is no `[[key]]` header left to render, and `_render_aot` looped over an empty body and emitted nothing — so the key vanished from the output entirely:

```python
doc = tomlkit.loads("[[a]]\nx = 1\n")
doc["a"].pop()
tomlkit.dumps(doc)   # '' — the 'a' key is gone
```

This renders the inline `a = []` form instead, which is what tomlkit already produces for an empty array built through the API (`doc["a"] = []`).

## The ordering part

Rendering `a = []` in body order is not enough. TOML only reads bare key/value pairs before the first table header, so an emptied array of tables sitting after one would be read back as a key *of that table*:

```toml
[t]
q = 2

a = []      # parses as t.a, not a
```

So empty arrays of tables are hoisted to just before the first table header. Non-empty ones are untouched and still render in place, byte for byte.

## Scope: the other half of the issue

The issue also covers `doc["x"]["y"].pop("z")` leaving an empty super table that renders as nothing. I have **not** fixed that here, because the current behaviour is pinned by two existing tests — `test_delete_out_of_order_table_key` and `test_overwriting_out_of_order_table` both assert that a super table emptied by `del` disappears from the output rather than rendering its own header.

Making emptied super tables render `[x.y]` breaks both. That looked like a deliberate call rather than an oversight, so I left it alone rather than rewriting your tests to suit. Happy to follow up if you'd like that changed — it is a small diff once the intended behaviour is settled.

## Tests

Four tests added, covering the bare case, the hoisting case, an emptied array preceded by a scalar, and a control asserting non-empty arrays of tables still round-trip unchanged.

Revert-verified: with `container.py` reverted, the first three fail and the control still passes.

Full suite on Windows / Python 3.12: **1055 passed** (1051 on a clean tree, plus these 4). `ruff check` and `ruff format --check` are clean on both changed files.

### DIFF
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -1666,3 +1666,44 @@ a.b = 1
     doc["z"] = 2
 
     assert doc.as_string() == "a.b = 1\nz = 2\n"
+
+
+def test_emptied_array_of_tables_renders_as_empty_array() -> None:
+    # https://github.com/python-poetry/tomlkit/issues/553
+    # An array of tables with no elements left has no `[[key]]` header to
+    # render, so it must fall back to the inline `key = []` form. Rendering
+    # nothing dropped the key entirely.
+    doc = parse("[[a]]\nx = 1\n")
+    doc["a"].pop()
+
+    assert doc.as_string() == "a = []\n"
+    assert parse(doc.as_string()) == {"a": []}
+
+
+def test_emptied_array_of_tables_is_hoisted_above_table_headers() -> None:
+    # https://github.com/python-poetry/tomlkit/issues/553
+    # TOML only reads bare key/value pairs before the first table header, so
+    # the inline fallback has to be emitted there rather than in body order.
+    # Left in place it would be parsed back as a key of the preceding table.
+    doc = parse("[t]\nq = 2\n\n[[a]]\nx = 1\n")
+    doc["a"].pop()
+
+    assert parse(doc.as_string()) == {"t": {"q": 2}, "a": []}
+    assert doc.as_string().index("a = []") < doc.as_string().index("[t]")
+
+
+def test_emptied_array_of_tables_keeps_preceding_scalars() -> None:
+    # https://github.com/python-poetry/tomlkit/issues/553
+    doc = parse("v = 9\n\n[[a]]\nx = 1\n")
+    doc["a"].pop()
+
+    assert parse(doc.as_string()) == {"v": 9, "a": []}
+
+
+def test_non_empty_array_of_tables_is_not_hoisted() -> None:
+    # https://github.com/python-poetry/tomlkit/issues/553
+    # Only emptied arrays of tables change form; ordinary ones must round-trip
+    # byte for byte.
+    content = "[t]\nq = 2\n\n[[a]]\nx = 1\n"
+
+    assert parse(content).as_string() == content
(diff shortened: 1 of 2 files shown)


davidpavlovschi (changes requested):
The empty fallback breaks nested arrays of tables because it keeps the full header prefix even though the fallback now renders as a key/value inside the current table scope. Minimal case: parse [t] followed by [[t.a]], pop the only t.a element, then serialize. This head emits [t] / t.a = [], which reparses as t.t.a instead of t.a. The same duplication occurs at deeper tables and inside an AoT element. Please add a nested regression that asserts doc.unwrap() equals parse(doc.as_string()).unwrap(). For an empty AoT rendered inside [t], the key/value fallback needs to be a = [] in that scope, not t.a = []. I ran the full suite on this head as well: 1,054 passed, 1 skipped; the current tests cover only root-level empty AoTs.

dchaudhari7177:
Thanks — you were right, and the case was wider than the one you posted.

Your minimal case (`[t]` then `[[t.a]]`, pop the only element) was fixed in `84ad727`. But writing the parametrized round-trip you asked for turned up two more shapes that were still wrong, both where the array's parent is an *implicit* super table:

| input (pop the only element) | before | reparsed as |
| --- | --- | --- |
| `[[t.a]]` | `a = []` | `{'a': []}` — the `t` is gone |
| `[t]` + `[[t.u.a]]` | `[t]` / `t.u.a = []` | `{'t': {'t': {'u': {'a': []}}}}` |

The first is your bug in the other direction: dropping the prefix is only right when the parent actually emitted a `[t]` to be bare inside. The second shows the prefix has to be *relative to the nearest emitted header*, not the absolute path.

So the fallbacks of a scope are now collected up front — walking through the super tables that emit no header of their own — and each is keyed relative to the scope that will read it. `[[t.a]]` renders `t.a = []`; `[[t.u.a]]` under `[t]` renders `u.a = []`.

Verification on this head:

- `doc.unwrap() == parse(doc.as_string()).unwrap()` over 32 generated shapes (explicit/implicit parent × one/two levels × sibling sub-table × preceding header × trailing header): all pass. On `84ad727` six of those failed.
- The parametrized regression now covers 12 shapes, plus two exact-output tests pinning `t.a = []` and `u.a = []`.
- Full suite: 1069 passed.
- Every `tests/examples/*.toml` still reparses byte-identical, so untouched documents are unaffected.

Jalst (approved):
LGTM! Verified that the changes requested by @davidpavlovschi in the previous review have been cleanly addressed in 84ad727.

### Verification Details
- Confirmed that prefix calculation now correctly accounts for the nearest emitted header scope rather than always using the absolute path:
  - Inside an explicit table (`[t]` then `[[t.a]]`), popping the only element yields `[t] / a = []` (instead of invalid `t.a = []`).
  - Inside a nested path (`[t]` then `[[t.u.a]]`), popping yields `[t] / u.a = []`.
  - When the parent table is implicit without an explicit header (`[[t.a]]` alone), popping yields `t.a = []`.
- All cases reparse to the exact matching unwrapped dictionary (`doc.unwrap() == parse(dumps(doc)).unwrap()`).
- Test suite: **1,069 passed** (including the 12 parametrized shape variations).
- Formatting and lint checks pass.

PR #568, opened 2026-07-20
Fix sub-table of a dotted key escaping its parent header path (#557)

### LINKED ISSUE #557: table escapes its parent
```python
#!/usr/bin/env python3

import tomlkit

doc = tomlkit.loads("""
[t]
a.b = 1
""")
doc["t"]["a"]["c"] = {}

output = tomlkit.dumps(doc)

print(output)
```
result
```
[t]
a.b = 1

[a.c]
```
in which `[a.c]` is not a child of `t` as it should be

earfman:
Fixes #557.

### Problem
Assigning a sub-table under a dotted-key super-table nested inside a real table caused the sub-table header to lose its parent path on render:

```python
doc = tomlkit.loads("\n[t]\na.b = 1\n")
doc["t"]["a"]["c"] = {}
tomlkit.dumps(doc)
# [t]
# a.b = 1
#
# [a.c]   <- escapes [t]; `c` is lost from t.a and reappears at the document root
```

`parse(dumps(doc)) != doc`, so this is a round-trip / data-loss bug. The same happened inside array-of-tables elements (`[[servers]]` produced `[a.c]` instead of `[servers.a.c]`).

### Cause
In `Container._render_table`, `prefix` served two purposes at once: the *relative* dotted-key prefix used for leaf items (so `a.b` stays inline under `[t]`) and the *absolute* header path used for nested sub-tables (which need `[t.a.c]`). The "dotted key inside table" branch dropped the prefix entirely — right for the leaf, wrong for a real sub-table nested underneath, which then lost the `t.` parent path. `_render_aot_table` had the same issue.

### Fix
Track the absolute header path separately via a new `header_prefix` argument. Leaf items continue to render off the relative key; nested sub-table and array-of-tables headers render off the absolute path. When a table carries a `display_name` (from a parsed explicit header) the two paths coincide, so existing output is unchanged.

### Tests
Added regression tests covering the table case, a chained dotted parent (`a.b.x` — the leaf stays relative while the sub-table header is absolute), and the array-of-tables case. The full suite passes (1053 tests).

---
[verified by Coretexa](https://coretexa.dev)

*Found and fixed with an integrity-first pipeline; the fix was independently verified by a separate agent than the one that found it before submission.*

### DIFF
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -1666,3 +1666,50 @@ a.b = 1
     doc["z"] = 2
 
     assert doc.as_string() == "a.b = 1\nz = 2\n"
+
+
+def test_subtable_of_dotted_key_keeps_parent_header_path() -> None:
+    # https://github.com/python-poetry/tomlkit/issues/557
+    # A dotted-key super table nested inside a real table (`[t]` -> `a.b = 1`)
+    # renders its leaf inline, but a sub-table added under it must keep the full
+    # header path (`[t.a.c]`), not escape to the document root as `[a.c]`.
+    doc = parse("[t]\na.b = 1\n")
+    doc["t"]["a"]["c"] = {}
+
+    expected = """\
+[t]
+a.b = 1
+
+[t.a.c]
+"""
+
+    assert doc.as_string() == expected
+    assert parse(doc.as_string()) == {"t": {"a": {"b": 1, "c": {}}}}
+    # Idempotent: re-rendering the parsed result must not drift.
+    assert parse(doc.as_string()).as_string() == expected
+
+    # Chained dotted parent keeps leaves relative while the sub-table is absolute.
+    doc = parse("[t]\na.b.x = 1\n")
+    doc["t"]["a"]["b"]["c"] = {}
+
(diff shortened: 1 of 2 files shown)


earfman:
Adding the **Agent Drafting Metadata** section from the PR template — I missed it when I opened this, and it shouldn't be left blank on a project that asks the question directly. Apologies for the omission.

## Agent Drafting Metadata

- **Agent:** Claude (Anthropic), run under my direction
- **Model:** claude-opus-5
- **Notes:** The agent produced the diagnosis, the patch and the regression tests; I reviewed and approved before submission and am accountable for the content. The root cause in `Container._render_table` was established by reading the renderer rather than by trial and error, and the fix was checked by reverting `container.py` while keeping the new tests, which fails them — so the tests genuinely gate the change rather than passing alongside it. Happy to walk through any part of it.

davidpavlovschi (approved):
Separating the relative leaf prefix from the absolute header prefix fixes the escaped sub-table path without changing dotted leaf rendering. I ran the complete suite on this head (1,052 passed, 1 skipped) and an additional 4,000 generated mutations across nested dotted keys under tables and array-of-table elements; every serialized document reparsed to the same unwrapped structure.

PR #570, opened 2026-07-22
test: add coverage for tomlkit._types wrapper operators

darrenhuai:
## Summary

`tomlkit/_types.py` was sitting at 60% coverage. The uncovered part is `_CustomList.__add__`/`__iadd__`, `_CustomDict.__or__`/`__ior__`, and the `wrap_method` helper — the machinery every `Array` and table type relies on to get list/dict operator support. None of it had a direct test.

While writing these I noticed something worth calling out: `Array + [...]` returns a plain `list`, not an `Array` (the `.copy()` in `__add__` doesn't preserve the subclass), whereas `Table | {...}` does come back as a `Table`. That's existing behavior, not something I changed — I just added a test that pins it down instead of leaving it implicit. Worth a look if that asymmetry wasn't intentional, but out of scope for a coverage PR.

`wrap_method` isn't called anywhere in the current codebase but is exported in `__all__`, so I tested it directly against a small stand-in class.

`_types.py` coverage: 60% → 100%. Full suite (1057 tests) still green, ruff and mypy --strict both clean.

## Agent Drafting Metadata

- Agent: Claude Code
- Model: claude-sonnet-5
- Notes: I used Claude Code to find the coverage gap and draft these tests. Ran the full test suite, ruff, and mypy myself before pushing, and read through the diff.

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -1,0 +1,83 @@
+from typing import Any
+
+import tomlkit
+
+from tomlkit._types import wrap_method
+from tomlkit.items import Array
+from tomlkit.items import Table
+
+
+def test_custom_list_add_returns_plain_list_with_combined_items() -> None:
+    arr = tomlkit.array()
+    arr.extend([1, 2, 3])
+
+    result = arr + [4, 5]  # noqa: RUF005 (exercising __add__ itself)
+
+    assert result == [1, 2, 3, 4, 5]
+    assert not isinstance(result, Array)
+
+
+def test_custom_list_iadd_mutates_in_place_and_keeps_wrapper_type() -> None:
+    arr = tomlkit.array()
+    arr.extend([1, 2])
+    original_id = id(arr)
+
+    arr += [9]
+
+    assert arr == [1, 2, 9]
+    assert isinstance(arr, Array)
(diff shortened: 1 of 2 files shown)


dimbleby:
`wrap_method` should have been removed at #460, better to remove it now than test it

davidpavlovschi (changes requested):
The operator coverage is useful, and the `__iadd__`, `__or__`, and `__ior__` cases match the implementations.

Two changes before this should merge:

1. Please remove `wrap_method` and its `__all__` entry instead of adding tests for it, as noted above. It has no callers and appears to be leftover from #460.
2. This is framed as a coverage-only PR, but `assert not isinstance(result, Array)` permanently specifies the wrapper-dropping behavior that the description itself flags as potentially unintended. Please either establish that a plain `list` is the intended API before pinning it, or keep the regression test focused on the combined values without asserting the result type.

That keeps the PR from turning two unresolved implementation details into supported behavior solely to reach 100% line coverage.

davidpavlovschi (approved):
Both points addressed at 4e9df46: wrap_method removed, and the __add__ test no longer pins the 
  wrapper type. CI green. Thanks for the quick turnaround.

PR #572, opened 2026-07-23
Fix validation of nested out-of-order tables

### LINKED ISSUE #571: Regression in 0.15.1: valid out-of-order child table raises KeyAlreadyPresent
## Summary

Tomlkit 0.15.1 rejects a valid document when a table contains an out-of-order
child, its concrete parent is declared afterward, and another table separates a
later sibling child.

This is a regression: `tomllib` and Tomlkit 0.13.3, 0.14.0, and 0.15.0 all
accept the document.

## Minimal reproduction

```python
import tomlkit

source = """\
[tool.ruff]
[tool.ruff.lint.a]
[tool.ruff.lint]
[[tool.poetry.source]]
[tool.ruff.lint.b]
"""

tomlkit.loads(source)
```

## Actual result

```text
tomlkit.exceptions.ParseError: Key "lint" already exists. at line 5 col 0
```

## Expected result

The document parses successfully, matching `tomllib`:

```python
{
    "tool": {
        "ruff": {"lint": {"a": {}, "b": {}}},
        "poetry": {"source": [{}]},
    }
}
```

Each header in the minimized example is necessary to trigger the failure. The
array-of-tables is not special: replacing `[[tool.poetry.source]]` with any
intervening table header also reproduces it, while removing the intervening
header makes it pass.

## Regression boundary

Reproduced on Python 3.13.8 and 3.14.2:

- stdlib `tomllib`: passes
- Tomlkit 0.13.3: passes
- Tomlkit 0.14.0: passes
- Tomlkit 0.15.0: passes
- Tomlkit 0.15.1 and current `master`: fail

## Suspected root cause

The regression appears to come from the concrete+super validation added in
#530 / commit d3c76f0.

When `[tool.ruff.lint.b]` is appended, the existing `lint` entry is represented
by `OutOfOrderTableProxy`, while the candidate is a `Table`.
`_validate_table_candidate()` checks:

```python
isinstance(existing, (Table, AoT)) != isinstance(v, (Table, AoT))
```

The proxy is table-like but fails the left-hand `isinstance` check, so valid
input is reported as a type conflict and raises `KeyAlreadyPresent("lint")`.

satriadhikara:
## Summary

- validate each raw out-of-order table fragment instead of materializing an
  `OutOfOrderTableProxy` during concrete/super-table validation
- add a regression test for a valid nested out-of-order table separated by an
  unrelated array-of-tables

The proxy was incorrectly treated as a non-table, causing valid TOML to raise
`KeyAlreadyPresent`. Validating the underlying `Table` fragments preserves the
existing collision checks without special-casing the proxy.

Closes #571.

## Checks

- `python -m pytest -q` — 1052 passed
- `ruff check tomlkit tests --exclude tests/toml-test`
- `ruff format --check tomlkit tests --exclude tests/toml-test`

The repository's strict mypy run still reports five pre-existing errors in
untouched `source.py`, `items.py`, and `api.py`.

## Agent Drafting Metadata

- Agent: OpenAI Codex
- Model: GPT-5.6 Sol
- Notes: Codex reproduced and minimized the regression, drafted the fix and
  tests, and prepared the issue and PR text.

### DIFF
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -643,6 +643,13 @@ def test_valid_out_of_order_independent_tables() -> None:
     assert doc.as_string() == "[a]\nx=1\n[zz]\n[a.b]\nc=1\n"
 
 
+def test_valid_nested_out_of_order_table_after_unrelated_table() -> None:
+    content = "[a]\n[a.b.c]\n[a.b]\n[[zz]]\n[a.b.d]\n"
+    doc = parse(content)
+    assert doc.unwrap() == {"a": {"b": {"c": {}, "d": {}}}, "zz": [{}]}
+    assert doc.as_string() == content
+
+
 def test_set_value_on_out_of_order_table_with_empty_concrete_part() -> None:
     # A super table defined after its sub-table (the "defining a super-table
     # afterward is ok" spec example) leaves an empty concrete `[x]` part.
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -423,19 +423,24 @@ class Container(_CustomDict):  # type: ignore[type-arg]
                 continue
 
             if k in current.value._map:
-                existing = current.value.item(k)
-                if isinstance(existing, (Table, AoT)) != isinstance(v, (Table, AoT)):
-                    raise KeyAlreadyPresent(k)
-                if k.is_dotted():
-                    raise TOMLKitError("Redefinition of an existing table")
-                if isinstance(existing, Table) and isinstance(v, Table):
-                    if not existing.is_super_table() and not v.is_super_table():
-                        # Both sides are concrete `[table]` definitions of the
-                        # same name; the table is declared twice.
+                index = current.value._map[k]
+                indices = index if isinstance(index, tuple) else (index,)
+                for i in indices:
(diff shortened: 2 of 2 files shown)


davidpavlovschi (approved):
Reviewed the validation path against the reported nested out-of-order case and the existing duplicate/redefinition checks. Iterating the raw fragments avoids treating the proxy as a non-table while still validating every fragment against the candidate. I also ran the complete suite locally on this head: 1,051 passed, 1 skipped. The focused regression preserves both the unwrapped structure and exact serialization.

satriadhikara:
Hi, following up on this PR after reading the discussion in #596.

I’ve updated the branch against current master and added regression tests for collisions across nested out-of-order fragments. All 1,058 tests pass, and lint and formatting checks pass.

Compared with #592, this approach applies the same type-conflict, concrete-table redefinition, and recursive checks to every stored fragment. This keeps validation in one path, avoiding proxy construction and a separate proxy-specific branch that would need to stay consistent with those checks. Both implementations pass the focused tests; I prefer this approach because it keeps the validation logic simpler.

Would you be able to take another look? Happy to address any concerns with this approach.

PR #575, opened 2026-07-28
Fix dotted-key child replacement capturing following siblings

### LINKED ISSUE #556: updating a table captures too many keys

shleder:
## Summary

Fixes #556.

When a dotted-key child such as `b` in `a.b = 1` is replaced with a table or array of tables through an out-of-order table proxy, that part begins rendering an `[a.b]` or `[[a.b]]` header at its original body position.

Dotted siblings that follow it, such as `a.c` and `a.d`, then fall inside that table's scope, corrupting the document when serialized output is parsed again.

This change detects the transition from an inline value to a table or AoT and moves the affected part after its inline same-group siblings, while keeping it before any subsequent same-group header part.

Because moving a body entry shifts other entries as well, the change rebuilds the complete key-to-index map and out-of-order metadata. This keeps unrelated plain keys, dotted groups, and other out-of-order tables correctly addressable and mutable. Proxy-derived state is rebuilt from the unchanged table parts after reordering.

Whitespace and trivia are preserved from the source document; the change does not add or remove blank lines.

Regression coverage includes:

- replacement with a non-empty table;
- replacement with an empty table;
- replacement with an array of tables;
- exact serialization and semantic round-trip;
- subsequent edits through the proxy;
- interleaved plain keys;
- interleaved dotted groups;
- empty-table and AoT interleaving.

Validation:

- focused #556/interleaving tests — 11 passed;
- `tests/test_toml_document.py` — 78 passed;
- complete suite including the `toml-test` corpus — 1058 passed;
- Ruff 0.15.21 — all checks passed;
- Ruff formatting — passed;
- mypy 1.19.1 — no errors attributable to the changed files;
- Python compilation — passed;
- `git diff --check` — passed.

## Agent Drafting Metadata

- Agent: Claude Code orchestration using `sonnet` for architecture, review and test execution, and `haiku` as the sole code-author role
- Model: `sonnet` is locally configured to route to Kimi K3; `haiku` is locally configured to route to Gemini 3.6 Flash
- Notes: These identities reflect locally verified routing configuration, not backend attestation. I reviewed the generated changes and validation evidence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [Unreleased]
+
+### Fixed
+
+- Replacing a dotted-key child with a table or AoT no longer captures following dotted siblings on round-trip. ([#556](https://github.com/python-poetry/tomlkit/issues/556))
+
 ## [0.15.1] - 2026-07-17
 
 ### Changed
--- a/tests/test_toml_document.py
+++ b/tests/test_toml_document.py
@@ -1326,6 +1326,26 @@ a.c = 2
     assert parse(doc.as_string()).unwrap() == expected_data
 
 
+@pytest.mark.parametrize(
+    ("replacement", "expected_output", "expected_value"),
+    [
+        ({"x": 9}, "q.c = 2\n\n[a.b]\nx = 9\n", {"x": 9}),
+        ({}, "q.c = 2\n\n[a.b]\n", {}),
+        ([{"x": 9}], "q.c = 2\n\n[[a.b]]\nx = 9\n", [{"x": 9}]),
+    ],
+)
+def test_replace_dotted_key_child_keeps_later_unrelated_dotted_key(
+    replacement: Any, expected_output: str, expected_value: Any
+) -> None:
+    doc = parse("a.b = 1\nq.c = 2\n")
+    doc["a"]["b"] = replacement
+
+    assert doc.as_string() == expected_output
+    expected_data = {"q": {"c": 2}, "a": {"b": expected_value}}
+    assert doc.unwrap() == expected_data
+    assert parse(doc.as_string()).unwrap() == expected_data
+
+
 def test_replace_with_comment() -> None:
     content = 'a = "1"'
     doc = parse(content)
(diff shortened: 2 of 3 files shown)


davidpavlovschi (changes requested):
The reorder still misses later inline dotted assignments from a different top-level group. A newly rendered table/AoT header captures those assignments when there is no later same-group sibling to push it past.

Minimal reproduction on this head:

```python
from tomlkit import parse

doc = parse("a.b = 1\nq.c = 2\n")
doc["a"]["b"] = {"x": 9}
output = doc.as_string()

assert doc.unwrap() == parse(output).unwrap()
```

`output` is currently `"\n[a.b]\nx = 9\nq.c = 2\n"`, so reparsing makes `q.c` a child of `a.b`. The same failure occurs with an AoT replacement. The move boundary needs to account for every later root-level inline dotted assignment that a rendered header would capture, not only entries whose `k.key` matches `group_key.key`. Please add a regression with only an unrelated group after the replaced entry; the current interleaving test also has a later `a.c`, which masks this case.

shleder:
Fixed the requested unrelated-group boundary case in 6f91002af6a9c0cdc40cb38c611d3506cabbdc23.

The exact `a.b = 1` / `q.c = 2` reproduction now serializes `q.c` before the new `[a.b]` header and preserves:

`doc.unwrap() == parse(output).unwrap()`

Added exact table, empty-table, and AoT regressions.

Validation:
- focused document tests: 81 passed;
- complete suite including initialized toml-test corpus: 1061 passed;
- focused Ruff check and format check passed;
- compilation passed;
- `git diff --check` passed.

Full-tree Ruff reports only pre-existing style findings in the vendored `toml-test` submodule.

PR #578, opened 2026-07-28
Fix out-of-order AoT element extensions

### LINKED ISSUE #577: unwrap raises KeyAlreadyPresent for an out-of-order AoT element extension
## Description

`tomlkit.parse()` accepts and exactly round-trips this valid document, but accessing `doc["y"]` or calling `doc.unwrap()` raises `KeyAlreadyPresent: Key "a" already exists`:

```python
from tomlkit import parse

content = """\
[[y.a]]
[b.d.x]
[[y.a.c]]
"""

doc = parse(content)
assert doc.as_string() == content
doc.unwrap()  # KeyAlreadyPresent: Key "a" already exists
```

Python's standard-library `tomllib` reads it as:

```python
{"y": {"a": [{"c": [{}]}]}, "b": {"d": {"x": {}}}}
```

The later `[[y.a.c]]` header extends the last element of the existing `y.a` array of tables. This works when the parts are adjacent, but an unrelated table makes `y` out of order. `OutOfOrderTableProxy` then sees `a` first as an `AoT` and later as an implicit `Table`; its existing fragment merge handles `AoT` + `AoT`, but not this `AoT` + super-table extension.

The example was initially flagged by @chuenchen309 in the description of #565. I reproduced it on current `master` and narrowed it down to the proxy merge path.

## Environment

- tomlkit 0.15.1 / current `master`
- Python 3.12

davidpavlovschi:
## Summary

Fixes #577.

An array-of-tables element can be extended by a later nested table header:

```toml
[[y.a]]
name = "first"

[unrelated]

[[y.a.c]]
kind = "child"
```

This document is valid and `parse()` already preserves it byte-for-byte. Because the unrelated table splits `y` into out-of-order parts, accessing `doc["y"]` or calling `doc.unwrap()` currently tries to append the later implicit `a` table beside the earlier `a` AoT and raises `KeyAlreadyPresent`.

`OutOfOrderTableProxy._merge_aot_fragment()` now handles this AoT + super-table combination by presenting the last AoT element through a nested proxy spanning both live source tables. It does not merge either source in place, so the original headers keep rendering in their original locations. Reads and `unwrap()` see the combined element, while edits route back to the correct source part.

Starting a subsequent `[[y.a]]` element after that nested extension is also valid, but currently fails during parsing because the new fragment is merged only against the last out-of-order part. Once a key is already out of order, the parser now preserves each later super-table as another part and lets the complete proxy/validation path merge it against all prior parts.

The regression test verifies:

- exact round-tripping before mutation;
- the unwrapped result against the `tomllib` interpretation;
- parsing and merging a subsequent AoT element after the nested extension;
- write-through for an original element key, a key inside the later nested AoT, and a newly added key;
- valid round-tripping after those mutations.

Validation:

- `1052 passed`
- 96 generated valid permutations matched `tomllib`, round-tripped exactly, and survived live mutation
- Ruff check and format check pass
- `git diff --check` passes

## Agent Drafting Metadata

- Agent: OpenAI Codex
- Model: GPT-5 family Codex
- Notes: David authorized the contribution. I reproduced the report against `tomllib`, traced the failure to the out-of-order proxy merge, implemented the live-view fix, and ran the complete test suite and formatting checks before opening this PR.


Jalst (approved):
LGTM! Verified the fix for #577 and tested read/write routing across fragments.

### Review Details
- **Mechanism**: Wrapping the extended last AoT element and its nested super-table extension in an `OutOfOrderTableProxy` cleanly preserves source table isolation without mutating either in place. This avoids duplicate header rendering on serialization while presenting a unified view to `__getitem__` and `unwrap()`.
- **Validation verified locally**:
  - Full test suite: **1,052 passed**.
  - Verified unwrapping matches `tomllib` output for out-of-order AoT element extensions.
  - Verified write-through mutations: updates to keys in earlier headers, nested child headers, and newly appended keys route back to their appropriate source tables without corrupting the rendered structure on `dumps()`.
  - Byte-for-byte round trip remains intact on unmutated documents.

PR #592, opened 2026-08-27
Fix spurious KeyAlreadyPresent on valid out-of-order child table (#571)

### LINKED ISSUE #571: Regression in 0.15.1: valid out-of-order child table raises KeyAlreadyPresent

chiliec:
Fixes #571.

## Problem

A regression in 0.15.1 rejects a valid document when a table has an out-of-order child, its concrete parent is declared afterwards, and another table separates a later sibling child:

```python
import tomlkit

tomlkit.loads("""\
[tool.ruff]
[tool.ruff.lint.a]
[tool.ruff.lint]
[[tool.poetry.source]]
[tool.ruff.lint.b]
""")
# tomlkit.exceptions.ParseError: Key "lint" already exists. at line 5 col 0
```

`tomllib` and tomlkit 0.13.3 / 0.14.0 / 0.15.0 all accept it.

## Root cause

The concrete+super validation added in 0.15.1 (`_validate_table_candidate`) treats only `Table`/`AoT` as table-like:

```python
if isinstance(existing, (Table, AoT)) != isinstance(v, (Table, AoT)):
    raise KeyAlreadyPresent(k)
```

When `lint` is spread across out-of-order headers, `current.value.item("lint")` returns an `OutOfOrderTableProxy`, not a `Table`. So the left side is `False` while the candidate `[tool.ruff.lint.b]` is a `Table` (`False != True`), and valid input is reported as a type conflict.

## Fix

- Recognise `OutOfOrderTableProxy` as table-like in the type-conflict check.
- When the existing entry is a proxy, recurse into each of its concrete `Table` fragments (`existing._tables`) so a genuine duplicate is still detected, rather than rejecting outright.

Real duplicates — including out-of-order ones like `[a.b]` / `[a]` / `[a.b]` — remain rejected.

## Test

Adds `test_parse_accepts_out_of_order_child_with_intervening_table` (the issue's exact document; asserts the parsed structure matches `tomllib` and the round-trip is byte-for-byte preserved). Also adds a CHANGELOG entry.

## Verification (local, Python 3.11)

- **RED→GREEN**: with the fix reverted the new test fails (`Key "lint" already exists`); with it, it passes.
- `pytest tests/` — **1053 passed, 0 failed** (full suite incl. the BurntSushi `toml-test` conformance corpus), no regressions.
- The parsed result equals `tomllib.loads(...)` and `dumps(doc)` reproduces the input exactly.
- `ruff check` introduces no new lint errors in the changed regions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [Unreleased]
+
+### Fixed
+
+- Fix a regression (since 0.15.1) where a valid out-of-order child table was rejected with `KeyAlreadyPresent` when its concrete parent was declared afterwards and another table separated a later sibling child. The existing entry is an `OutOfOrderTableProxy`, which was not recognised as table-like during table-candidate validation. ([#571](https://github.com/python-poetry/tomlkit/issues/571))
+
 ## [0.15.1] - 2026-07-17
 
 ### Changed
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -480,6 +480,30 @@ def test_parse_rejects_collisions_across_out_of_order_fragments(
         parse(content)
 
 
+def test_parse_accepts_out_of_order_child_with_intervening_table() -> None:
+    """A valid out-of-order child table must not be rejected just because its
+    concrete parent is declared afterwards and another table separates a later
+    sibling child. Regression for #571 (broken in 0.15.1): the existing entry
+    is an OutOfOrderTableProxy, which was not recognised as table-like and
+    raised a spurious KeyAlreadyPresent."""
+    content = (
+        "[tool.ruff]\n"
+        "[tool.ruff.lint.a]\n"
+        "[tool.ruff.lint]\n"
+        "[[tool.poetry.source]]\n"
+        "[tool.ruff.lint.b]\n"
+    )
+    doc = parse(content)
+    assert doc.unwrap() == {
+        "tool": {
+            "ruff": {"lint": {"a": {}, "b": {}}},
+            "poetry": {"source": [{}]},
+        }
+    }
+    # Round-trip must be preserved exactly.
+    assert dumps(doc) == content
+
+
 def test_create_super_table_with_table() -> None:
     data = {"foo": {"bar": {"a": 1}}}
     assert dumps(data) == "[foo.bar]\na = 1\n"
--- a/tomlkit/container.py
+++ b/tomlkit/container.py
@@ -428,7 +428,9 @@ class Container(_CustomDict):  # type: ignore[type-arg]
                 # rather than a Table. It is still table-like, so treat it as a
                 # table both for the type-conflict check and when recursing into
                 # the subtree (see #571).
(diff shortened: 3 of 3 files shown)


Jalst (approved):
LGTM! Verified the fix for #571 and tested for potential duplicate detection regressions.

### Findings
- **Root cause**: In `_validate_table_candidate`, `current.value.item(k)` returns an `OutOfOrderTableProxy` when table segments are separated by other tables. Because `OutOfOrderTableProxy` inherits from `_CustomDict` rather than `Table`/`AoT`, the type parity check `isinstance(existing, (Table, AoT)) != isinstance(v, (Table, AoT))` previously evaluated to `False != True`, incorrectly rejecting valid out-of-order children with a spurious `KeyAlreadyPresent`.
- **Fix verification**:
  1. Including `OutOfOrderTableProxy` in table-like recognition restores compliance with `tomllib` and earlier tomlkit behavior.
  2. Recursing candidate validation into each fragment (`for fragment in existing._tables: self._validate_table_candidate(fragment, v)`) ensures genuine duplicate table declarations (e.g. out-of-order redefinitions) and scalar collisions continue to be caught as parse errors.
- **Test suite**:
  - Full suite passes: **1,053 passed**.
  - Document round-tripping and `unwrap()` match standard `tomllib` output byte-for-byte.

PR #606, opened 2026-09-18
fix: keep add_line comments from swallowing the closing bracket

### LINKED ISSUE #580: `Array.add_line()` doesn't serialize as documented (or expected) and can even produce syntactically invalid TOML
*Update: This seems to be part of a larger problem; see [my comment below](https://github.com/python-poetry/tomlkit/issues/580#issuecomment-5194300758) for additional context.*

If an `Array` isn't multiline, then calling its `.add_line` method doesn't add a trailing line separator. When the added line includes a comment, this causes `.as_string` and similar to produce invalid TOML:

```python
>>> import tomlkit
>>> array = tomlkit.array()
>>> array.add_line("foo", comment="bar")
>>> print(array.as_string())
[
    "foo", # bar]
>>> doc = tomlkit.document()
>>> doc.add("array", array)
{'array': ['foo']}
>>> print(doc.as_string())
array = [
    "foo", # bar]

>>> tomlkit.loads(doc.as_string())
  File "<python-input-7>", line 1, in <module>
    tomlkit.loads(doc.as_string())
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/api.py", line 52, in loads
    return parse(string)
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/api.py", line 99, in parse
    return Parser(string).parse()
           ~~~~~~~~~~~~~~~~~~~~^^
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/parser.py", line 139, in parse
    item = self._parse_item()
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/parser.py", line 238, in _parse_item
    return self._parse_key_value(True)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/parser.py", line 331, in _parse_key_value
    val = self._parse_value()
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/parser.py", line 435, in _parse_value
    return self._parse_array()
           ~~~~~~~~~~~~~~~~~^^
  File "/home/five35/.local/share/hatch/env/virtual/python-scripts/S8f9twao/python-scripts/lib/python3.14/site-packages/tomlkit/parser.py", line 612, in _parse_array
    raise self.parse_error(UnexpectedCharError, self._current)
tomlkit.exceptions.UnexpectedCharError: Unexpected character: '\x00' at line 2 col 0
```

I think the simplest fix might be to make calling `.add_line` implicitly change single-line arrays to multiline. If nothing else, it doesn't make semantic sense to "add a line" to something which can't contain multiple lines, so hopefully that behavior wouldn't be surprising to anyone.

00200200:
Fixes #580

`Array.add_line(..., comment=...)` builds a compact array. The last item is a `#` comment with an empty trail, so `as_string()` glued the closing `]` onto that comment:

```
[
    "foo", # bar]
```

That is not valid TOML. If the last compact item is a comment without a newline, a newline is inserted before `]`. The existing `add_line` tests that already end with a newline are unchanged.

## Agent Drafting Metadata
- Agent: Cursor
- Model: Grok 4.6
- Notes: Helped write the regression test and the compact-render guard. I reproduced the invalid TOML on main first, then applied the fix.

**Tests**
```
pytest tests/test_items.py -k add_line
```


00200200:
Closing this — I missed the existing PRs already covering #580. Sorry for the noise.

PR #607, opened 2026-09-18
chore: Define project metadata via [project] table (PEP 621)

hroncok:
## Summary

This leads to a small METADATA diff:

     Name: tomlkit
     Version: 0.15.1
     Summary: Style preserving TOML library
    -License: MIT
    +License-Expression: MIT
     License-File: LICENSE
     Author: Sébastien Eustace
     Author-email: sebastien@eustace.io
     Requires-Python: >=3.9
    -Classifier: License :: OSI Approved :: MIT License
     Classifier: Programming Language :: Python :: 3
     Classifier: Programming Language :: Python :: 3.9

The difference is only wrt. license, PEP 639.

The required poetry-core version was bumped to support this.

There is no immediate need for this change,
it's pure modernization.
My personal motivation is for me to make it easier to experiment with different build backends.

## Agent Drafting Metadata

- Agent: Claude
- Model: Opus 5
- Notes: I used LLM to draft this change, then I reviewed it line by line and changed whitespace slightly. I asked for build artefacts before and after the change and to note their results. I verified those differences myself.


hroncok:
The pre-commit fix is not related to my changes at all.

pyproject.toml
```diff
@@ -1,28 +1,28 @@
...
   { path = "tomlkit/py.typed" },
   { path = "tests", format = "sdist" },
   { path = "docs", format = "sdist" },
   { path = "CHANGELOG.md", format = "sdist" },
 ]
 
-[tool.poetry.dependencies]
-python = ">=3.9"
-
 [tool.poetry.group.dev.dependencies]
```
dimbleby, line 26: could move these to `[dependency-groups]` too while you are at it

hroncok: I could. But I don't use poetry and I do not know how to integrate it into other workflows.

dimbleby: I don't see what that's got to do with it - the suggestion was that, like the other changes you are making, you could move details out of poetry-specific sections and into standard sections that other tools will recognise.

But I see that my comment crossed with frostming merging the PR anyway, and it's certainly not super-important.

hroncok: I assumed poetry groups are different to standardized dependency groups. If they are equivalent, I can send a follow-up PR.

dimbleby: They are equivalent

hroncok: https://github.com/python-poetry/tomlkit/pull/609

